PR #3, opened 2022-08-19
added hello solana program using seahorse lang

cdhiraj40:
Added a simple hello Solana program written in python using [seahorse lang](https://seahorse-lang.org/)


jpcaulfi (approved):
Looks good!

PR #5, opened 2022-09-10
Add direct calling to the System program for creating accounts

dadepo:


dadepo:
Close in favor of #13

PR #9, opened 2022-09-11
Improvements to checking accounts

dadepo:
This PR 
- sets the Program ID
- Adds a README.

The Program ID is required because if it is not set, the tests passes but shows no tests were run:

```


  0 passing (0ms)

✨  Done in 0.70s.

```

This is because the `describe` function, is annotated to be an `async` function. Removing the `async` keyword and running the test, the issue becomes obvious, and it fails with the following error:

```
ts-mocha -p ./tsconfig.json -t 1000000 ./tests/test.ts

Error: Invalid public key input
    at new PublicKey... 
```


ngundotra:
Tagging @realbuffalojoe for review

ngundotra:
But also, please see feedback on #10 about your README updates to `create-account` example

realbuffalojoe (approved):
lgtm

PR #13, opened 2022-09-15
Add direct call to system program for creating accounts

dadepo:
Extend the create account test to include creating an account by direct invocation of the System program


ngundotra:
I think it would be helpful to modify this example to print out the created account address, so curious developers can run this on devnet + quickly inspect the new accounts.

cc @realbuffalojoe

dadepo:
> See Noah's comment above.

I thought that was addressed in [this](https://github.com/solana-developers/program-examples/pull/13/commits/e1623de4e8a65324def0f5c80dc9e18d9fb96eac) commit. Or I misinterpreted? and what needs to be updated is different?

realbuffalojoe:
Yeah, I think it was just in another one of your PRs. Looks good

realbuffalojoe (approved):
lgtm

PR #15, opened 2022-09-18
Add readme program derived addresses

dadepo:


ngundotra:
Can you change the README to more closely follow what the docs say? For example something like:

> Program derived addresses allow programmatically generated signatures to be used when [calling between programs](https://docs.solana.com/developing/programming-model/calling-between-programs#cross-program-invocations).

> Using a program derived address, a program may be given the authority over an account and later transfer that authority to another. This is possible because the program can act as the signer in the transaction that gives authority.

ngundotra:
Can you remove your logging statements from the `counter` program?

Once done, tag @realbuffalojoe for review

realbuffalojoe:
Added comments to your README. Also agree with Noah please omit the logs. Ty

dadepo:
@ngundotra @realbuffalojoe updated as suggested.

realbuffalojoe:
See README comments

dadepo:
> See README comments

This [here](https://github.com/solana-developers/program-examples/pull/15#issuecomment-1291428244) is the README comments I can see. And I already updated the README as suggested. Could it be I am missing the comments you referring to? As I can't see it

realbuffalojoe (changes requested):

realbuffalojoe:
Lmk if you can see it now

PR #16, opened 2022-09-25
Fix running counter anchor

dadepo:
Tried running the Anchor version of the Anchor program and saw it was failing due to hardcoded user path to the wallet. This PR fixes that


ngundotra:
Great thank yiu!

PR #20, opened 2023-03-04
Change create.rs

yeni111828:
Error in create.rs during build program.
Changed "system_program::" in line 24, 26 to "anchor_lang::system_program::"


yeni111828 (commented):
ready to use

PR #25, opened 2023-04-28
Add missing dependencies for cNFT vault

Woody4618:
There were some dependencies missing, so the tests wouldn't run. 

Now they run, but the tests are still failing for me. 
image

Maybe @loopcreativeandy you could have another look at it when you have time?


buffalojoec (approved):
Thanks Jonas! I will merge to add dependencies and have a look as well

Eianaftyio (approved):
.diff

Eianaftyio:
.diff

PR #31, opened 2023-06-18
Added examples to create merkle tree for compressed nfts

akshatcoder-hash:


buffalojoec:
Closing since this isn't a program example and we're already discussing with `web3-examples`

PR #37, opened 2023-07-28
Added Seahorse create token program

akshaydhayal:


akshaydhayal:
@valentinmadrid can you check this?

valentinmadrid:
Thanks for your pr, could you make sure to use a tilde instead of /users/username in Anchor.toml so that tests can be ran by any user please

PR #41, opened 2023-08-31
add rust workspace and update crates version

DaviRain-Su:


acheroncrypto:
Some programs use pretty old versions so it would be very helpful to get this merged. Amazing effort @DaviRain-Su 👏.

PR #58, opened 2023-12-05
Added nft_staking example on Anchor (both for points and for tokens)

L0STE:
Created a nft staking program in anchor both for points and for tokens


ZYJLiu:
Thank you for adding these examples! Could you please also include a README to provide an explanation for the two programs for readers?

L0STE:
Done, let me know if i have to change something or if it's good enough like this! (I added a pNFT version too since i think it's useful)

ZYJLiu:
Thanks for including the README!

- Can you double check that running `anchor build` and `anchor test` works for the programs included in the PR?
- Can you also add the `package.json` file for the `staking-for-tokens` project?

- It seems like there might be a dependency conflict. When I run `anchor build` I'm getting the following error.     
    Here is the related anchor release note regarding [mpl-token-metadata](https://github.com/coral-xyz/anchor/blob/master/docs/src/pages/docs/release-notes.md?plain=1#L287-L303) for anchor 0.29.

```
error[E0277]: the trait bound `T: borsh::de::BorshDeserialize` is not satisfied
   --> src/utils/mod.rs:176:21
    |
176 |     let result: T = try_from_slice_unchecked(data)?;
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^ the trait `borsh::de::BorshDeserialize` is not implemented for `T`
    |
note: required by a bound in `solana_program::borsh::try_from_slice_unchecked`
   --> src/borsh.rs:38:1
help: consider further restricting this bound
    |
167 | pub fn try_from_slice_checked<T: TokenMetadataAccount + borsh::de::BorshDeserialize>(
    |                                                       +++++++++++++++++++++++++++++
```

PR #74, opened 2024-04-03
crud-dapp

brimigs:


jacobcreech:
@brimigs could you remove the `.next` folder? App-wise this may make sense as its own repo.

brimigs:
> @brimigs could you remove the `.next` folder? App-wise this may make sense as its own repo. The `program-examples` repo doesn't have any ui in it today

Moved over to here: https://github.com/solana-developers/dApp-examples/pull/1

PR #78, opened 2024-05-06
Update to Anchor 0.30.0 and add Token Extensions examples

ZYJLiu:
- update basics and token examples to Anchor 0.30.0
- add anchor examples for most token extensions (excluding group/member extensions and confidential transfers because they are not enabled yet)
- cnft and pyth examples not updated, due to dependency issues


heyAyushh:
We shouldn't remove pnpm lockfiles as per this document https://pnpm.io/git

PR #79, opened 2024-05-08
fix solana-program versions, edited workflows

heyAyushh:
- bumped versions in basic native programs
- anchor 0.30.0 works with solana 1.18.8 and above


Elgayar777 (approved):
- [ ]

PR #81, opened 2024-06-02
tokens native, formatter, trigger action on project changes

heyAyushh:
- Added linting and formatting with [biomejs](http://biomejs.dev), this is added as git prehook, means at pre commit biome will format and lint files.
- GItHub actions now find and build 'anchor' and 'native' projects, So we don't need to maintain the list of projects to build in those specific actions
- Added [.ghaignore](https://github.com/heyAyushh/program-examples/blob/main/.github/.ghaignore) file to ignore projects to build or test
- setup-anchor github action, now also caches anchor binaries
- Anchor building and testing action detects the changes in a pr and only builds on those projects. 
Reducing workflow run to 5 min * n projects from building all programs listed (2 hours)

- Added tests for token 2022 native in solana-bankrun programs
```
tokens/token-2022/mint-close-authority/native
tokens/token-2022/non-transferable/native
tokens/token-2022/default-account-state/native
tokens/token-2022/transfer-fee/native
tokens/token-2022/multiple-extensions/native
```


jacobcreech:
Was every file changed?

jacobcreech (approved):

PR #84, opened 2024-06-12
Add isTransferring check to transfer hook examples

Woody4618:
This check checks the source token account if it is currently transferring to make sure that the transfer hook can only be called during a transfer. 
This is to make sure that for example a counter or similar can not be manipulated by calling the transfer hook directly.


ZYJLiu (approved):
chane

Woody4618:
Thank you :)

PR #86, opened 2024-06-12
Fix all the biome errors

mikemaccana:
`npm run lint` (ie `pnpx @biomejs/biome lint ./`) now passes.


deleteme.log
heyAyushh, line 1: Just this

heyAyushh (commented):
Looks good to me 🫡

heyAyushh (approved):

PR #106, opened 2024-07-26
refactor actions, included programs from .ghaignore

heyAyushh:
Do squash and merge.

Some projects were ignored when Stable solana was breaking because of dependency issue, 
removing the names so that they can be built again.

### DIFF
--- a/.github/.ghaignore
+++ b/.github/.ghaignore
@@ -18,13 +18,6 @@ tokens/spl-token-minter/anchor
 tokens/token-swap/anchor
 tokens/transfer-tokens/anchor
 
-# not building with stable solana
-tokens/token-2022/mint-close-authority/native
-tokens/token-2022/non-transferable/native
-tokens/token-2022/default-account-state/native
-tokens/token-2022/transfer-fee/native
-tokens/token-2022/multiple-extensions/native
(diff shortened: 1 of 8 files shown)


heyAyushh:
Anchor builds and tests are skipping due to a logical error. 🚧

heyAyushh:
Refactor Anchor Action and Solana Action

- [ ] Combine build and test for projects (build only once and test them once, instead of building twice and testing once)
       (downside of this is parallel building and testing) 
- [ ] instead of creating a array of projects and then use regex to find, create regex earlier as a string itself
- [ ] Batch array of projects into multiple jobs considering 256 jobs limit, (mike suggested using projects in matrix earlier)
- [ ] Use shell functions (repeating code)
- [ ] Use error messages (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message
- [ ] Use Job Summary (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary)
- [ ] update token solana crates to 1.18.17
- [ ] update anchor ones to 0.30.1 (there's a warning in 0.30.0 projects).

PR #109, opened 2024-08-02
Skip ignoring spl2.0 dependency issue (#2)

heyAyushh:
Refactor Anchor Action and Solana Action

- [x] Combine build and test for projects (build only once and test them once, instead of building twice and testing once)
       (downside of this is parallel building and testing) 
- [x] instead of creating a array of projects and then use regex to find, create regex earlier as a string itself
- [x] Batch array of projects into multiple jobs considering 256 jobs limit, (mike suggested using projects in matrix earlier)
- [x] Use shell functions (repeating code)
- [x] Use error messages (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message
- [x] Use Job Summary (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary)
- [x] update projects that use spl-tokens, with solana crates to 1.18.17
- [ ] ~~update anchor ones to 0.30.1 (there's a warning in 0.30.0 projects).~~ [opened issue to track this](https://github.com/solana-developers/program-examples/issues/110),

### DIFF
--- a/.github/.ghaignore
+++ b/.github/.ghaignore
@@ -18,23 +18,19 @@ tokens/spl-token-minter/anchor
 tokens/token-swap/anchor
 tokens/transfer-tokens/anchor
 
-# not building with stable solana
-tokens/token-2022/mint-close-authority/native
-tokens/token-2022/non-transferable/native
-tokens/token-2022/default-account-state/native
-tokens/token-2022/transfer-fee/native
-tokens/token-2022/multiple-extensions/native
-
 # not building
 oracles/pyth/anchor
+
+# avm broke hence can't build
+# can't use anchor.toml [toolchain] with stable rust
+# update these programs to use latest anchor version to fix
+tokens/token-2022/group/anchor
+tokens/token-2022/immutable-owner/anchor
+tokens/token-2022/interest-bearing/anchor
+tokens/token-2022/memo-transfer/anchor
 
 # not building
 compression/cutils/anchor
 compression/cnft-vault/anchor
 # builds but need to test on localhost
 compression/cnft-burn/anchor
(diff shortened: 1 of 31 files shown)


.github/.ghaignore
```diff
@@ -18,23 +18,19 @@ tokens/spl-token-minter/anchor
...
-tokens/token-2022/default-account-state/native
-tokens/token-2022/transfer-fee/native
-tokens/token-2022/multiple-extensions/native
-
 # not building
 oracles/pyth/anchor
 
+# avm broke hence can't build
+# can't use anchor.toml [toolchain] with stable rust
+# update these programs to use latest anchor version to fix
```
mikemaccana, line 26:
```suggestion
# TODO: update these programs to use latest anchor version to fix
```

Add a TODO for this, people sometimes search for `TODO` and fix them for you!

.github/.ghaignore
```diff
@@ -18,23 +18,19 @@ tokens/spl-token-minter/anchor
...
 
-# not building with stable solana
-tokens/token-2022/mint-close-authority/native
-tokens/token-2022/non-transferable/native
-tokens/token-2022/default-account-state/native
-tokens/token-2022/transfer-fee/native
-tokens/token-2022/multiple-extensions/native
-
 # not building
 oracles/pyth/anchor
```
mikemaccana, line 22:
```suggestion
# not building. TODO: fix.
oracles/pyth/anchor
```

mikemaccana (approved):
Approved!

jacobcreech (approved):

PR #112, opened 2024-08-06
Transfer hook : token stop

tiago18c:
Example on how to use a transfer hook to enable / disable token transfers while storing the current transferability state in the mint metadata


tokens/token-2022/transfer-hook/token-stop/anchor/programs/token-stop/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+        Ok(())
+    }
+
+    // we dont care about the token amount transferred
+    #[interface(spl_transfer_hook_interface::execute)]
+    pub fn execute(ctx: Context<Execute>/* , amount: u64 */) -> Result<()> {
+        let mint_info = ctx.accounts.mint.to_account_info();
+        let mint_data = mint_info.data.borrow();
+        let mint = StateWithExtensions::<Token2022Mint>::unpack(&mint_data)?;
+
```
Woody4618, line 79: You could add a check for good pratice to check if its actually currently transferring 
fn check_is_transferring(ctx: &Context<TransferHook>) -> Result<()> {
    let source_token_info = ctx.accounts.source_token.to_account_info();
    let mut account_data_ref: RefMut<&mut [u8]> = source_token_info.try_borrow_mut_data()?;
    let mut account = PodStateWithExtensionsMut::<PodAccount>::unpack(*account_data_ref)?;
    let account_extension = account.get_extension_mut::<TransferHookAccount>()?;

    if !bool::from(account_extension.transferring) {
        return err!(TransferError::IsNotCurrentlyTransferring);
    }

    Ok(())
}

tokens/token-2022/transfer-hook/token-stop/anchor/programs/token-stop/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    TransfersDisabled,
+}
+
+#[program]
+pub mod token_stop {
+    use anchor_spl::token_interface::spl_token_metadata_interface::state::Field;
+
+    use super::*;
+
+    pub fn stop_transfer(ctx: Context<StopTransfer>) -> Result<()> {
```
Woody4618, line 37: Looks like anyone could trigger the stop or resume the transfer. A nice addition would be to save the authority also in the token meta data or add an admin key check or use the mint authority as admin .

Woody4618 (approved):
Super nice usecase \o/ Love the idea of having the config in the meta data.

PR #116, opened 2024-08-23
Fix slow warnings

mikemaccana:
## Before 

Screenshot 2024-08-23 at 12 40 59

## After 

Screenshot 2024-08-23 at 12 41 25

@heyAyushh you might want to do this elsewhere if you like.

If you want, you can not merge this, and use `--slow` instead: https://mochajs.org/#-slow-ms-s-ms


heyAyushh (commented):
When anchor-bankrun will be added, 
not using `--slow` will be good for comparison on how much time each txn is taking.

PR #132, opened 2024-10-18
feat: steel counter

thearyanag:
- Add counter in [steel framework](https://github.com/regolith-labs/steel)
- Add tests with solana-bankrun


thearyanag:
gm @heyAyushh , pls check now

PR #133, opened 2024-10-18
Add transfer sol example for poseidon

DogukanGun:
- Add Transfer Sol example for [poseidon](https://github.com/DogukanGun/program-examples/tree/main/tokens/transfer-tokens/poseidon)
- Add test case for example


DogukanGun:
> 5\. cargo clippy

Hi @heyAyushh , thank you for the review. Branch is updated according to your comment.

PR #160, opened 2024-10-20
Add basics/account-data/steel

robertohuertasm:
Adds an account data example using the [Steel framework](https://github.com/regolith-labs/steel) containing both [Bankrun](https://kevinheavey.github.io/solana-bankrun/) tests and Rust tests.

Update to steel 2.1


ZYJLiu:
thank you for the contribution, please add the example to the steel repo instead where the steel team is better positioned to maintain the example

PR #161, opened 2024-10-20
Add basics/close-account/poseidon

adpthegreat:
- added poseidon close_account program to basics/close-program folder
- updated dependencies to use latest versions of anchor
- used pnpm in place of yarn, , replaced all instances of yarn with pnpm
- wrote solana bankrun tests 
- ran pnpm fix
Screenshot 2024-10-29 at 11 22 14


Yangatech (commented):
Instead of using u8 for the name field, consider using a String or a fixed-size byte array for more practical usage. Ensure the allocated account space matches the actual size needed for the UserAccount struct. Add logging with msg! to track key actions like account creation and closure.

basics/close-account/poseidon/close_account_program/ts-programs/src/close_account_program.ts
Yangatech, line 1: storing the name as a u8 may limit usability; switching to a string or byte array would be better.

Yangatech (changes requested):

Yangatech:
Ohh okay thank you so much for your feedback

PR #230, opened 2024-10-26
Add basics/favorites/anchor - added get_favourites fn & tested.

Praharx:
tests added for solana bankrun and anchor tests for newly added get_favourites fn.


Praharx:
I'll be adding this PR through different branch.

PR #280, opened 2024-10-29
Add /tokens/poseidon/create_token

adpthegreat:
- added poseidon create_token program to tokens/create_token folder
- updated dependencies to use latest versions of anchor
- used pnpm in place of yarn, , replaced all instances of yarn with pnpm
- wrote solana bankrun tests ran pnpm fix


adpthegreat:
Poseidon does not currently support creating tokens, so I'm currently implementing create_metadata_v3 in poseidon to be able to create_tokens , and will make a PR to the poseidon repo when I'm done

adpthegreat:
Okay so i've implemented create_metadata_v3 and created a PR for it in the poseidon repo and poseidon-types repo will update you when its merged

PR #330, opened 2024-11-23
Fix calculations in token-swap anchor example

dsctt:
* Fix liquidity ratio calculation in deposit and add tests
Liquidity pool ratio was being calculated by multiplying token supplies instead of dividing causing incorrect amounts of attempted deposits and liquidity token mints. This bug wasn't caught by tests because the ratio is only used on deposits subsequent to the initial deposit and the tests previously only covered the initial deposit.

* Fix invariant validation calculation in swap and add tests
A suspected typo in the invariant calculation used for validation in `swap_exact_tokens_for_tokens.rs` would cause the program to fail even if the swap should otherwise be successful. This bug wasn't caught by tests because they did not cover the case when initial deposit contains amounts where a < b, which is the case when the failure occurs.

* Anchor version
Using anchor 0.30.0 gave me this error which is why this PR has 0.29.0.
```
**error****: custom attribute panicked**                                                                                   
  **-->** programs/token-swap/src/lib.rs:13:1                                                                            
   **|** **13** **|** #[program]                                                                                                      
   **|** **^^^^^^^^^^**                                                                                                      
   **|**                                                                                                                 
   **=** **help**: message: Safety checks failed: Failed to get program path
```
(Upgrading to anchor 0.30.1 also gives the same error but with a help message of `**help**: message: compiler/fallback mismatch #949`). I couldn't figure this out in a reasonable amount of time. If someone knows how to resolve this that would be great. If it's imperative that we use 0.30, I can try to spend more time on it.

### DIFF
--- a/tokens/token-swap/anchor/Anchor.toml
+++ b/tokens/token-swap/anchor/Anchor.toml
@@ -4,6 +4,9 @@
 resolution = true
 skip-lint = false
 
+[programs.localnet]
+swap_example = "AsGVFxWqEn8icRBFQApxJe68x3r9zvfSbmiEzYFATGYn"
+
 [programs.devnet]
 swap_example = "AsGVFxWqEn8icRBFQApxJe68x3r9zvfSbmiEzYFATGYn"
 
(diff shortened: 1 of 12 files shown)


nlgripto (approved):

nlgripto:
@buffalojoec 🫡

dyodx:
oops. this was a while ago, i can redo this if you guys want the fix

PR #365, opened 2025-03-24
Fix/rewrite steel learning example

ubadineke:
Rewrote the entire learning example to align with the current Steel implementation.
- Refactored for clarity and correctness
- Verified implementation with working code and tests


ubadineke:
@Perelyn-sama check this out

Perelyn-sama:
I'll review later tonight, thanks for putting this together ❤️

tokens/create-token/steel/api/src/state/mod.rs
Perelyn-sama, line 1: You don't need the state mod in this example, especially the `Token` type 

Add the token fields in api::instruction::CreateToken like so: 
```
pub struct CreateToken {
    pub name: [u8; 32],
    pub symbol: [u8; 8],
    pub uri: [u8; 128],
    pub decimals: u8,
}
```

since you're removing the state mod, you'd need get the metadata pda pubkey in api::sdk:: create_token function. This would work:
```
  let metadata_pda = Pubkey::find_program_address(
        &[METADATA, mpl_token_metadata::ID.as_ref(), mint.as_ref()],
        &mpl_token_metadata::ID,
    );
```

Perelyn-sama:
The rust code isn't formatted, you can fix that with `cargo fmt`. 

You also need to add bankrun tests to get this merged here and there might be other things particular to this codebase you need to add so check the contributing guide

ubadineke:
@Perelyn-sama I've made the changes, please review.

Perelyn-sama:
LGTM 

cc @jacobcreech @heyAyushh

jacobcreech (approved):

PR #423, opened 2025-10-01
fix ci errors

Perelyn-sama:


basics/rent/anchor/programs/rent-example/src/lib.rs
```diff
@@ -1,4 +1,6 @@
 #![allow(clippy::result_large_err)]
+#![allow(unexpected_cfgs)]
+#![allow(deprecated)]
```
jacobcreech: These are everywhere. Why? Can't we get rid of these in anchor 0.32.1?

Perelyn-sama: I added it before the new release and there was no other way to fix it then. I'll update it now.

PR #504, opened 2026-01-05
feat(anchor/escrow): Cancel offer

a-hariti:
Adds a `cancel_offer` instruction to the Anchor escrow program along with tests following the same patterns.

**Note**: I made sure the tests pass with the changes I made in this PR #502


a-hariti:
But this ix makes so much sense and the program logic feels incomplete without it

PR #538, opened 2026-02-24
add hello solana asm example

Perelyn-sama:


dev-jodee (changes requested):
Couple of small things to change :) thx for the work

dev-jodee (commented):
re on the solana dependency added -> Yes, but that dependency isn't used anywhere in your code itself

dev-jodee (approved):

PR #567, opened 2026-04-14
fix(token-swap): correct ratio calculation and liquidity assertion in deposit_liquidity

### LINKED ISSUE #512: Bug: token-swap Anchor program
The `token-swap` anchor program has a logic bug,

specifically a in the `deposit_liquidity` instruction when it calculates the LP tokens to award the depositor based on the pool tokens ratio.

It should be `ratio = Pool token A / Pool Token B`, currently it's doing a multiplication:

https://github.com/solana-developers/program-examples/blob/d7fcf9949ba7919805a0cc75a78d7a34d139db6f/tokens/token-swap/anchor/programs/token-swap/src/instructions/deposit_liquidity.rs#L40-L42

I have prepared a solution in #505

HkSolDev:
## Summary

Fixes the `deposit_liquidity` instruction which had incorrect ratio
calculation logic and test assertions that didn't catch the bug.

## Root Cause

The ratio was computed with `checked_mul` (pool_a × pool_b) instead of
`checked_div` (pool_a / pool_b), so deposit proportionality was wrong.
Tests were also passing the wrong token amount for the B slot, masking
the bug entirely.

### `programs/token-swap/src/instructions/deposit_liquidity.rs`
- Changed ratio from `pool_a.checked_mul(pool_b)` → `pool_a.checked_div(pool_b)`
  so proportionality is calculated correctly
- Replaced `.unwrap()` with `.ok_or(DepositTooSmall)?` on `checked_div`
  to avoid panic when pool_b is zero

### `tests/deposit-liquidity.ts`
- Pass `depositAmountB` (1M) instead of `depositAmountA` (4M) for token B slot
  so unequal deposits are actually tested
- Fix liquidity assertion: use `sqrt(A × B) - minLiquidity` formula
  instead of `A - minLiquidity` which only worked when A == B
- Fix token B balance assertion: subtract `depositAmountB` not `depositAmountA`

## Testing

\```bash
cargo build-sbf  # ✅ passes (warnings only)
anchor test      # ✅ 7/7 tests pass
\```

Closes #512

Screenshot 2026-04-14 at 10 53 40 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1802,12 +1802,12 @@ name = "hello-solana-asm-program"
 version = "0.1.0"
 dependencies = [
  "litesvm",
- "solana-instruction 3.0.0",
+ "solana-instruction 3.3.0",
  "solana-keypair",
  "solana-native-token 3.0.0",
  "solana-pubkey 3.0.0",
  "solana-system-interface 2.0.0",
- "solana-transaction",
+ "solana-transaction 3.1.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


HkSolDev:
@claude

PR #595, opened 2026-06-16
add pinocchio transfer-tokens example

dev-codecraft:
## What

Adds a Pinocchio implementation of `tokens/transfer-tokens`, covering the core SPL flow:

1. **CreateToken** — create and initialize an SPL token mint
2. **MintTokens** — create the recipient's associated token account (idempotent) and mint to it
3. **TransferTokens** — transfer tokens between associated token accounts

This is the first Pinocchio example under `tokens/`, following the pattern of the recently merged pinocchio examples.

## How

The program dispatches three instructions by a leading discriminator byte (`0`=CreateToken, `1`=MintTokens, `2`=TransferTokens), using `pinocchio-token`, `pinocchio-associated-token-account`, and `pinocchio-system`.

A `solana-bankrun` test exercises the full **create → mint 150 → transfer 50** flow and asserts the resulting **100 / 50** balances. The legacy SPL Token and Associated Token Account programs are bundled with bankrun, so no extra fixtures are loaded.

The `package.json` / `pnpm-lock.yaml` / `tsconfig.json` are copied from `basics/repository-layout/pinocchio` so `--frozen-lockfile` passes (only deps already present in the lockfile are used).

## Checks run locally

- `cargo fmt --check` ✅
- `cargo clippy -- -D warnings` ✅
- `biome check` ✅

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2282,6 +2282,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "pinocchio-associated-token-account"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0bd92823a97fb327d7509dfd7cbfa3ead56e9fc0e131972bc0e28ab7036be31a"
+dependencies = [
+ "solana-account-view",
+ "solana-address 2.6.0",
+ "solana-instruction-view",
+ "solana-program-error",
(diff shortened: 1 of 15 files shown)


dev-codecraft:
Closing — opened from the wrong account, will reopen from the correct one.

PR #599, opened 2026-06-21
add pinocchio create-token example

MarkFeder:
Adds a Pinocchio implementation of `tokens/create-token`, continuing the Pinocchio token-examples lane after `transfer-tokens` (#596) and `escrow` (#598).

## What it does
Creates an SPL Token mint and attaches an on-chain Metaplex metadata account (name, symbol, URI) — the same flow as the existing `anchor` and `native` options.

## Notes
- The Metaplex Token Metadata program has no typed Pinocchio crate, so the `CreateMetadataAccountV3` CPI is built by hand (raw `InstructionView` / `cpi::invoke`), matching the on-chain wire format.
- Single instruction, no discriminator byte — the instruction data is Borsh `[name, symbol, uri, decimals]`, identical to the `native` example's wire format.
- The bankrun test loads the Metaplex program from a mainnet-dumped `token_metadata.so` fixture via a `postinstall` script, the same approach as the `anchor` example.
- The mint authority is passed as a non-signer and aliased to the payer (which signs), mirroring the `native` example.

## Files
- New: `tokens/create-token/pinocchio/` (program + bankrun test)
- Edited: root `Cargo.toml` (workspace member), `README.md` (pinocchio link), `Cargo.lock`


tokens/create-token/pinocchio/program/src/instructions/create_token.rs
```diff
@@ -0,0 +1,131 @@
...
+    let metadata_data = build_metadata_data(&args);
+    let metadata_accounts = [
+        InstructionAccount::writable(metadata_account.address()),
+        InstructionAccount::readonly(mint_account.address()),
+        InstructionAccount::readonly_signer(mint_authority.address()),
+        InstructionAccount::writable_signer(payer.address()),
+        // Update authority — recorded only, not required to sign for V3.
+        InstructionAccount::readonly(mint_authority.address()),
+        InstructionAccount::readonly(system_program.address()),
+    ];
```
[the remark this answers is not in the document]
MarkFeder: The update authority is intentionally `readonly` (non-signer) here — `CreateMetadataAccountV3` only requires the mint authority and payer to sign; the update-authority account is recorded but not required to sign for V3 (noted in the inline comment at this line). The native helper's `update_authority_is_signer: true` is an optional convention that Metaplex V3 does not enforce, so `readonly` is the minimal-correct meta, and the full CPI is exercised on-chain by the bankrun test in CI. Happy to flip it to `readonly_signer` for closer parity with the native example if you'd prefer that for teaching clarity.

tokens/create-token/pinocchio/program/src/instructions/create_token.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    log!("Creating metadata account");
+    let metadata_data = build_metadata_data(&args);
+    let metadata_accounts = [
+        InstructionAccount::writable(metadata_account.address()),
+        InstructionAccount::readonly(mint_account.address()),
+        InstructionAccount::readonly_signer(mint_authority.address()),
+        InstructionAccount::writable_signer(payer.address()),
+        // Update authority — recorded only, not required to sign for V3.
+        InstructionAccount::readonly(mint_authority.address()),
```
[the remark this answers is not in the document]
MarkFeder: The update authority is intentionally `readonly` (non-signer) here — `CreateMetadataAccountV3` only requires the mint authority and payer to sign; the update-authority account is recorded but not required to sign for V3 (noted in the inline comment at this line). The native helper's `update_authority_is_signer: true` is an optional convention that Metaplex V3 does not enforce, so `readonly` is the minimal-correct meta, and the full CPI is exercised on-chain by the bankrun test in CI. Happy to flip it to `readonly_signer` for closer parity with the native example if you'd prefer that for teaching clarity.

tokens/create-token/pinocchio/tests/test.ts
```diff
@@ -0,0 +1,116 @@
...
+// Borsh schema for the instruction data, matching the program's `CreateTokenArgs`
+// (and the native example's wire format).
+const CreateTokenArgsSchema: borsh.Schema = {
+  struct: {
+    token_title: "string",
+    token_symbol: "string",
+    token_uri: "string",
+    token_decimals: "u8",
+  },
+};
```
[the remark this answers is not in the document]
MarkFeder: Right — Borsh is positional, so the field names don't affect the wire bytes and this is correct at runtime. The TS schema names (`token_title`/`token_symbol`/…) are intentionally descriptive labels for the test reader, while the Rust `CreateTokenArgs` uses the terser on-chain `name`/`symbol`/… . Since the two sit in different languages and are wire-compatible, I left the local naming as-is, but I can rename the TS schema to match the Rust fields 1:1 if you'd prefer the tighter cross-reference.

MarkFeder:

tokens/create-token/pinocchio/prepare.mjs
```diff
@@ -0,0 +1,35 @@
...
+    name: "token_metadata.so",
+  },
+];
+
+const outputDir = "tests/fixtures";
+
+try {
+  mkdirSync(outputDir, { recursive: true });
+  // Point the Solana CLI at mainnet, where the canonical program lives.
+  execSync("solana config set -um", { stdio: "inherit" });
```
dev-jodee: don't think we should change the config of the user's cli, program dump can take the url as a parameter, better do it then

tokens/create-token/pinocchio/program/src/instructions/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+            symbol,
+            uri,
+            decimals,
+        })
+    }
+}
+
+/// Reads a Borsh `string` (a 4-byte little-endian length prefix followed by that
+/// many UTF-8 bytes) starting at `*offset`, advancing `offset` past it.
+fn read_borsh_string<'a>(data: &'a [u8], offset: &mut usize) -> Result<&'a [u8], ProgramError> {
```
dev-jodee: this shoudl probably be in a util file instead of the mod.rs, would be cleaner

tokens/create-token/pinocchio/program/src/instructions/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+/// options share an identical wire format.
+pub struct CreateTokenArgs<'a> {
+    pub name: &'a [u8],
+    pub symbol: &'a [u8],
+    pub uri: &'a [u8],
+    pub decimals: u8,
+}
+
+impl<'a> CreateTokenArgs<'a> {
+    /// Parses the instruction data: three Borsh strings followed by a `u8`.
```
dev-jodee: think this should be in create_token.rs file as well since its related to that specific ix

tokens/create-token/pinocchio/program/src/instructions/mod.rs
```diff
@@ -0,0 +1,61 @@
+use pinocchio::error::ProgramError;
+
+mod create_token;
+
+pub use create_token::*;
+
+/// Size (in bytes) of an SPL Token mint account.
+pub const MINT_SIZE: usize = 82;
```
dev-jodee: Does pinocchio have that as a const already? might be able to reuse

tokens/create-token/pinocchio/tests/test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+// Derive the Metaplex metadata PDA for a mint.
+function getMetadataAddress(mint: PublicKey): PublicKey {
+  return PublicKey.findProgramAddressSync(
+    [Buffer.from("metadata"), TOKEN_METADATA_PROGRAM_ID.toBuffer(), mint.toBuffer()],
+    TOKEN_METADATA_PROGRAM_ID,
+  )[0];
+}
+
+describe("Create Token (Pinocchio)", async () => {
```
dev-jodee: I think this runs 0 test, since describe is async, it doesn't wait for anything ? All the setup (async) should be in before()

dev-jodee (changes requested):
thanks for the work, couple of comments !

tokens/create-token/pinocchio/prepare.mjs
MarkFeder: Good call — dropped the `solana config set -um` and now pass the cluster straight to the dump: `solana program dump -um <id> <file>`. No more touching the user's CLI config.

tokens/create-token/pinocchio/program/src/instructions/mod.rs
MarkFeder: Yes it does — switched to `pinocchio_token::state::Mint::LEN` and removed the hand-rolled `MINT_SIZE` const.

tokens/create-token/pinocchio/program/src/instructions/mod.rs
MarkFeder: Agreed — moved `CreateTokenArgs` (struct + `parse`) into `create_token.rs` since it's specific to that instruction. The Metaplex program-id const went with it.

tokens/create-token/pinocchio/program/src/instructions/mod.rs
MarkFeder: Done — the borsh string helpers (`read_borsh_string`/`push_borsh_string`) now live in a new `util.rs`, so `mod.rs` is just module wiring.

tokens/create-token/pinocchio/tests/test.ts
MarkFeder: Nice catch — you're right, the `async` `describe` callback registered the `it` blocks too late. Moved the async bankrun setup into a `before()` hook and kept `describe` synchronous.

MarkFeder:
@dev-jodee thanks for the review! Pushed a commit addressing all five points:

- `prepare.mjs` dumps via `solana program dump -um` instead of mutating the CLI config
- reuse `pinocchio_token::state::Mint::LEN` instead of a custom `MINT_SIZE`
- `CreateTokenArgs` moved into `create_token.rs`; borsh string helpers moved to a `util.rs`; `mod.rs` is now just wiring
- test setup moved into a `before()` hook so the `it` blocks actually run

Ready for another look when you have a moment.

tokens/create-token/pinocchio/program/src/instructions/create_token.rs
```diff
@@ -0,0 +1,176 @@
...
+};
+use pinocchio_log::log;
+use pinocchio_system::instructions::CreateAccount;
+use pinocchio_token::{instructions::InitializeMint2, state::Mint};
+
+use crate::instructions::util::{read_borsh_string, write_borsh_string, write_bytes};
+
+/// The Metaplex Token Metadata program ID
+/// (`metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s`).
+const TOKEN_METADATA_PROGRAM_ID: pinocchio::Address =
```
dev-jodee: nit: I'd put all metaplex related thing in a `util_metaplex.rs` or something like that, so you keep your instruction file cohesive

tokens/create-token/pinocchio/program/src/instructions/create_token.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    let args = CreateTokenArgs::parse(data)?;
+
+    // Rent-exempt minimum for the mint, computed with integer math using the
+    // network's default rent parameters (`DEFAULT_LAMPORTS_PER_BYTE` already
+    // folds in the 2-year exemption threshold). We avoid
+    // `Rent::try_minimum_balance`, whose floating-point exemption-threshold path
+    // emits an instruction the bankrun test VM rejects ("unsupported BPF
+    // instruction").
+    let lamports = (ACCOUNT_STORAGE_OVERHEAD + Mint::LEN as u64) * DEFAULT_LAMPORTS_PER_BYTE;
```
dev-jodee, line 81: rent calculation should be done via the Rent struct, like in this example https://github.com/solana-foundation/subscriptions/blob/e937688e644553c2d765bf4f0098318aa60a5be4/program/src/instructions/helpers/program.rs#L33

tokens/create-token/pinocchio/program/src/instructions/create_token.rs
```diff
@@ -0,0 +1,176 @@
...
+    log!("Initializing mint account");
+    InitializeMint2 {
+        mint: mint_account,
+        decimals: args.decimals,
+        mint_authority: mint_authority.address(),
+        freeze_authority: Some(mint_authority.address()),
+    }
+    .invoke()?;
+
+    log!("Creating metadata account");
```
dev-jodee, line 102: for an example it's good, but just FYI, log! functions will add some CU to your transaction :) but here it's fine

dev-jodee (changes requested):
Thanks for the fixes ! just a couple more :)

tokens/create-token/pinocchio/program/src/instructions/create_token.rs
MarkFeder: Thanks! I actually tried exactly this — `Rent::get()?.try_minimum_balance(Mint::LEN)?` — but it makes the bankrun test fail with `Program failed to complete ... unsupported BPF instruction`.

`try_minimum_balance`/`minimum_balance` compile the `f64` exemption-threshold branch, and solana-bankrun's VM rejects that float instruction at load time (verified: reverting to the `Rent` call gives 2 failing tests, reverting back gives 2 passing). pinocchio 0.10.2 doesn't expose the rent fields, so there's no way to read a live `Rent` and compute the minimum with integer math.

So I kept `(ACCOUNT_STORAGE_OVERHEAD + Mint::LEN) * DEFAULT_LAMPORTS_PER_BYTE` — still pinocchio's own rent constants, and `DEFAULT_LAMPORTS_PER_BYTE` (6960) already folds in the 2-year threshold, so it yields the identical lamport value without the float codegen. There's an inline comment in the code explaining this tradeoff.

Happy to switch to `Rent::get()` if you'd rather have that and are OK dropping the bankrun assertion that then fails — your call. 🙏

tokens/create-token/pinocchio/program/src/instructions/create_token.rs
MarkFeder: Done — moved the program ID, the CreateMetadataAccountV3 discriminator, the metadata-data size bound, and the instruction-data builder into a new `util_metaplex.rs` in 2027b84. `create_token.rs` now just imports them and stays focused on the account/CPI flow.

MarkFeder:
@dev-jodee thanks for the second pass! Pushed 2027b84:

- **Metaplex cohesion** — moved all Metaplex bits (program ID, discriminator, data-size bound, instruction-data builder) into a new `util_metaplex.rs`; `create_token.rs` is now just the account/CPI flow.
- **Rent** — replied inline: your suggestion is what I'd default to, but `try_minimum_balance` compiles an `f64` branch that solana-bankrun's VM rejects (verified it flips the tests red/green), and pinocchio 0.10.2 doesn't expose the rent fields for an integer path. Kept pinocchio's own `ACCOUNT_STORAGE_OVERHEAD`/`DEFAULT_LAMPORTS_PER_BYTE` constants (same lamport value). Glad to switch if you'd prefer `Rent::get()` and dropping that assertion.
- **log! CU** — good call, noted; kept them here since it's an example and clarity wins.

Rebuilt with cargo build-sbf + bankrun (2 passing), fmt/clippy clean.

dev-jodee (approved):

dev-jodee:
lgtm, just realized commits aren't signed :') so can't merge it, can you re-sign the commits ? will merge after thanks

MarkFeder:
@dev-jodee the branch is now re-signed — all commits show as Verified on GitHub. Ready to merge whenever you get a chance. Thanks!

dev-jodee:
thanks for the work @MarkFeder

PR #612, opened 2026-06-24
add pinocchio pda-mint-authority example

MarkFeder:
Adds a Pinocchio port of the `tokens/pda-mint-authority` example, alongside the existing `anchor` and `native` versions.

## What it does

A program-derived address — not a wallet — is the mint and freeze authority for every NFT this program creates. Three instructions, dispatched by a leading discriminator byte (matching the native `MyInstruction` enum):

- **Init** (`0`) — creates the mint-authority PDA (`[b"mint_authority"]`), signed by its own seeds, and persists the canonical bump in the account.
- **Create** (`1`) — creates a 0-decimal SPL mint whose authority is the PDA, then attaches a Metaplex metadata account via a hand-rolled `CreateMetadataAccountV3` CPI. The metadata CPI is authorized with the PDA's seeds via `invoke_signed`.
- **Mint** (`2`) — creates the payer's associated token account (idempotent), mints the single token, then creates the master edition via a hand-rolled `CreateMasterEditionV3` CPI (`max_supply = Some(1)`). Both the `MintTo` and master-edition CPIs are signed by the PDA.

The new building block here versus the other token examples is **PDA-as-signer**: `Init`, the metadata CPI, the `MintTo`, and the master-edition CPI all sign as the PDA using `pinocchio::cpi::{Seed, Signer}` and `invoke_signed`, rather than relying on a wallet signature. The bump recorded by `Init` is read back from the PDA account to rebuild the signer seeds without re-deriving the address on-chain.

## Tests

`tests/test.ts` runs under `solana-bankrun`, loading the program plus the Token Metadata program (dumped from mainnet into `tests/fixtures` by `prepare.mjs`). Three cases:
- Init asserts the PDA account is owned by the program and stores the expected bump.
- Create asserts the mint is owned by the Token program and the metadata account is owned by Token Metadata and contains the NFT name.
- Mint asserts the ATA holds exactly 1 token and the master edition account exists and is owned by Token Metadata (proving the PDA-signed `CreateMasterEditionV3` CPI succeeded).

### DIFF
--- a/tokens/pda-mint-authority/pinocchio/prepare.mjs
+++ b/tokens/pda-mint-authority/pinocchio/prepare.mjs
@@ -1,0 +1,35 @@
+// Dumps the Metaplex Token Metadata program from mainnet into the bankrun
+// fixtures directory so the test can load it into the local test validator.
+// Runs automatically via the `postinstall` script.
+//
+// Uses only the Node.js standard library (no extra dependencies). Errors are
+// logged but not fatal — a missing fixture will surface as a clear test failure
+// when bankrun can't find `token_metadata.so`.
+
+import { execSync } from "node:child_process";
+import { mkdirSync, rmSync } from "node:fs";
+import { join } from "node:path";
+
+const programs = [
+  {
+    id: "metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s",
+    name: "token_metadata.so",
+  },
+];
+
+const outputDir = "tests/fixtures";
+
+try {
+  mkdirSync(outputDir, { recursive: true });
+  // Point the Solana CLI at mainnet, where the canonical program lives.
+  execSync("solana config set -um", { stdio: "inherit" });
+
+  for (const { id, name } of programs) {
+    const outputFile = join(outputDir, name);
+    rmSync(outputFile, { force: true });
+    execSync(`solana program dump ${id} ${outputFile}`, { stdio: "inherit" });
+    console.log(`Dumped ${id} -> ${outputFile}`);
+  }
+} catch (error) {
+  console.error(`Failed to prepare program fixtures: ${error.message}`);
+}
--- a/tokens/pda-mint-authority/pinocchio/prepare.mjs
+++ b/tokens/pda-mint-authority/pinocchio/prepare.mjs
@@ -21,13 +21,13 @@ const outputDir = "tests/fixtures";
 
 try {
   mkdirSync(outputDir, { recursive: true });
-  // Point the Solana CLI at mainnet, where the canonical program lives.
-  execSync("solana config set -um", { stdio: "inherit" });
 
   for (const { id, name } of programs) {
     const outputFile = join(outputDir, name);
     rmSync(outputFile, { force: true });
-    execSync(`solana program dump ${id} ${outputFile}`, { stdio: "inherit" });
+    // `-um` points this dump at mainnet without mutating the developer's
+    // global Solana CLI cluster config.
+    execSync(`solana program dump -um ${id} ${outputFile}`, { stdio: "inherit" });
     console.log(`Dumped ${id} -> ${outputFile}`);
   }
 } catch (error) {
--- a/tokens/pda-mint-authority/pinocchio/prepare.mjs
+++ b/tokens/pda-mint-authority/pinocchio/prepare.mjs
@@ -6,30 +6,30 @@
 // logged but not fatal — a missing fixture will surface as a clear test failure
 // when bankrun can't find `token_metadata.so`.
 
-import { execSync } from "node:child_process";
-import { mkdirSync, rmSync } from "node:fs";
-import { join } from "node:path";
+import { execSync } from 'node:child_process';
+import { mkdirSync, rmSync } from 'node:fs';
+import { join } from 'node:path';
 
 const programs = [
-  {
-    id: "metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s",
-    name: "token_metadata.so",
-  },
+    {
+        id: 'metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s',
+        name: 'token_metadata.so',
+    },
 ];
 
-const outputDir = "tests/fixtures";
+const outputDir = 'tests/fixtures';
 
 try {
-  mkdirSync(outputDir, { recursive: true });
+    mkdirSync(outputDir, { recursive: true });
 
-  for (const { id, name } of programs) {
-    const outputFile = join(outputDir, name);
-    rmSync(outputFile, { force: true });
-    // `-um` points this dump at mainnet without mutating the developer's
-    // global Solana CLI cluster config.
-    execSync(`solana program dump -um ${id} ${outputFile}`, { stdio: "inherit" });
-    console.log(`Dumped ${id} -> ${outputFile}`);
-  }
+    for (const { id, name } of programs) {
+        const outputFile = join(outputDir, name);
+        rmSync(outputFile, { force: true });
+        // `-um` points this dump at mainnet without mutating the developer's
+        // global Solana CLI cluster config.
+        execSync(`solana program dump -um ${id} ${outputFile}`, { stdio: 'inherit' });
+        console.log(`Dumped ${id} -> ${outputFile}`);
+    }
 } catch (error) {
-  console.error(`Failed to prepare program fixtures: ${error.message}`);
+    console.error(`Failed to prepare program fixtures: ${error.message}`);
 }
--- a/tokens/pda-mint-authority/pinocchio/program/src/instructions/create.rs
+++ b/tokens/pda-mint-authority/pinocchio/program/src/instructions/create.rs
@@ -1,0 +1,156 @@
+use alloc::vec::Vec;
+
+use pinocchio::{
+    cpi::{invoke_signed, Seed, Signer},
+    error::ProgramError,
+    instruction::{InstructionAccount, InstructionView},
+    sysvars::{rent::Rent, Sysvar},
+    AccountView, Address, ProgramResult,
+};
+use pinocchio_log::log;
+use pinocchio_pubkey::derive_address;
+use pinocchio_system::instructions::CreateAccount;
+use pinocchio_token::instructions::InitializeMint2;
+
+use crate::instructions::{CreateTokenArgs, MINT_SIZE, TOKEN_DECIMALS, TOKEN_METADATA_PROGRAM_ID};
+use crate::state::MintAuthorityPda;
+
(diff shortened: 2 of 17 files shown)


tokens/pda-mint-authority/pinocchio/prepare.mjs
```diff
@@ -0,0 +1,35 @@
...
+    name: "token_metadata.so",
+  },
+];
+
+const outputDir = "tests/fixtures";
+
+try {
+  mkdirSync(outputDir, { recursive: true });
+  // Point the Solana CLI at mainnet, where the canonical program lives.
+  execSync("solana config set -um", { stdio: "inherit" });
```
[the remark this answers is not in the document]
MarkFeder: Good catch — fixed in 1cfe23d. `prepare.mjs` now dumps the program with a per-command `solana program dump -um …` instead of `solana config set -um`, so `pnpm install` no longer mutates the developer's global Solana CLI cluster.

tokens/pda-mint-authority/pinocchio/program/src/instructions/init.rs
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    let bump = *data.first().ok_or(ProgramError::InvalidInstructionData)?;
+
+    // Verify the supplied account is the canonical PDA for this bump.
+    let pda = derive_address(
+        &[MintAuthorityPda::SEED_PREFIX],
+        Some(bump),
+        program_id.as_array(),
+    );
```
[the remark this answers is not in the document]
MarkFeder: Leaving this as-is, for two reasons:

1. **Consistency with the established pinocchio pattern.** This is exactly what the already-merged `tokens/escrow/pinocchio` does (`make_offer.rs`): take the bump from instruction data, then `derive_address(&[seed], Some(bump), program_id)` and reject the tx unless the supplied account equals that PDA. Every pinocchio example in the repo follows this; matching it keeps the teaching examples uniform.

2. **Toolchain constraint.** Deriving the canonical bump on-chain needs `find_program_address`/`create_program_address`, whose off-target implementation in `solana-address` is gated behind the `curve25519` feature. CI lint runs `cargo clippy -- -D warnings` on the **host** target (no `--target sbf`), so referencing it there fails to compile unless we pull in the curve25519 dependency — which the lightweight pinocchio stack intentionally avoids. `derive_address` (from `pinocchio-pubkey`) is the host-compatible primitive, and it only derives for a given bump.

On safety: the supplied mint-authority account is validated against `derive_address(Some(bump))`, and `create`/`mint` re-derive the signer seeds from the bump persisted in that account, so the program is internally consistent. A client that deliberately passes a non-canonical bump only affects its own address derivation; the test (and any normal client) sources the bump from `findProgramAddressSync`, which always returns the canonical one.

MarkFeder:

dev-jodee (changes requested):
Will require changes based on https://github.com/solana-foundation/program-examples/pull/624 final version

MarkFeder:
Applied the `kit` + `litesvm` template from #624 here as well. The test now builds and signs transactions with `@solana/kit` and runs them on `litesvm`, so `@solana/web3.js` and `solana-bankrun` are dropped entirely. Pinned `@solana/kit` to `^6.10.0` (litesvm's kit major) for a single kit in the tree, and bumped `typescript`/`tsconfig` (es2022 + dom, `@types/node`) so the suite is type-clean under `tsc --noEmit`. Verified locally: `cargo build-sbf` + `ts-mocha` → **3 passing**, plus `tsc --noEmit` and `biome` clean. Commit is signed.

MarkFeder:
Applied the same refinements @dev-jodee asked for on #624 (now merged):

- **program**: rent is computed with `Rent::get()?.try_minimum_balance(..)?` in both the Init (PDA account) and Create (mint) instructions (the integer-math workaround was only needed to dodge the f64 opcode the old bankrun VM rejected; litesvm runs the real syscall).
- **test**: the token, associated-token and system program ids now come from the official `@solana-program/token` / `@solana-program/system` packages, and the minted amount is read with `getTokenDecoder().decode(...).amount` instead of a raw byte offset. Token Metadata has no official `@solana-program` client, so its id stays hand-rolled.
- **tsconfig**: `moduleResolution: bundler` for the packages' subpath exports.

`cargo build-sbf` + `ts-mocha` → **3 passing**; `tsc`, biome, fmt and clippy clean; frozen-lockfile OK. Commit is signed.

dev-jodee:
Still want this one. Heads up though — #656 changed the ground rules: pinocchio is 0.11 (workspace deps), and tests are mocha+tsx with `@solana/kit` + litesvm 1.x now. `tokens/token-2022/non-transferable/pinocchio` shows the current pattern, AGENTS.md has the rest. Rebase when you get a chance and I'll review.

MarkFeder:
@dev-jodee #624 is merged, so I've rebased this onto latest `main` (resolves the conflict) and migrated the program to **pinocchio 0.11** to match — it dropped the `pinocchio-pubkey` crate, so PDAs are now derived on-chain with `Address::find_program_address` (via `solana-address`'s `curve25519` feature, the same pattern the merged `block-list` example uses).

That migration also **resolves the earlier non-canonical-bump comment**: `init` now derives the canonical bump on-chain and rejects any mismatch, so a non-canonical bump can no longer create the account at a different address. (The toolchain constraint I'd cited for keeping the old approach no longer applies on 0.11.)

Host `cargo clippy -D warnings` and `cargo fmt --check` are clean. Ready for another look.

tokens/pda-mint-authority/pinocchio/program/src/instructions/create.rs
```diff
@@ -0,0 +1,134 @@
...
+    else {
+        return Err(ProgramError::NotEnoughAccountKeys);
+    };
+
+    let args = CreateTokenArgs::parse(data)?;
+
+    // Recover the PDA bump recorded by `init` and confirm the supplied account is
+    // the canonical mint-authority PDA.
+    let bump = MintAuthorityPda::deserialize(&mint_authority.try_borrow()?)?.bump;
+    let (pda, _) = Address::find_program_address(&[MintAuthorityPda::SEED_PREFIX], program_id);
```
dev-jodee: You can use create_program_address instead, provide the bump and then it doesnt need to find the pda address, it directly derives it

tokens/pda-mint-authority/pinocchio/program/src/instructions/mint.rs
```diff
@@ -0,0 +1,129 @@
...
+    let [mint_account, metadata_account, edition_account, mint_authority, associated_token_account, payer, system_program, token_program, _associated_token_program, _token_metadata_program] =
+        accounts
+    else {
+        return Err(ProgramError::NotEnoughAccountKeys);
+    };
+
+    // Recover the PDA bump recorded by `init` and confirm the supplied account is
+    // the canonical mint-authority PDA.
+    let bump = MintAuthorityPda::deserialize(&mint_authority.try_borrow()?)?.bump;
+    let (pda, _) = Address::find_program_address(&[MintAuthorityPda::SEED_PREFIX], program_id);
```
dev-jodee: You can use create_program_address instead, provide the bump and then it doesnt need to find the pda address, it directly derives it

tokens/pda-mint-authority/pinocchio/tests/test.ts
```diff
@@ -0,0 +1,221 @@
...
+const TOKEN_METADATA_PROGRAM_ID = address('metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s');
+
+// Instruction discriminators (the Borsh enum variant index).
+const INIT = 0;
+const CREATE = 1;
+const MINT = 2;
+
+// Borsh schema for the Create instruction data, matching the program's
+// `CreateTokenArgs` (and the native example's wire format).
+const CreateTokenArgsSchema: borsh.Schema = {
```
dev-jodee: follow what @amilz did here https://github.com/solana-foundation/program-examples/pull/675

tokens/pda-mint-authority/pinocchio/tests/test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+async function getMasterEditionAddress(mint: ReturnType<typeof address>) {
+    const [edition] = await getProgramDerivedAddress({
+        programAddress: TOKEN_METADATA_PROGRAM_ID,
+        seeds: ['metadata', addressEncoder.encode(TOKEN_METADATA_PROGRAM_ID), addressEncoder.encode(mint), 'edition'],
+    });
+    return edition;
+}
+
+async function getAssociatedTokenAddress(mint: ReturnType<typeof address>, owner: ReturnType<typeof address>) {
```
dev-jodee: pretty sure there's already built in functions ot get ATAs, Master edition pda, etc

dev-jodee (changes requested):

tokens/pda-mint-authority/pinocchio/program/src/instructions/create.rs
MarkFeder: Done in a500411c — `create` now derives the PDA directly with `create_program_address(&[SEED, &[bump]], program_id)`. The canonical bump is already persisted in the mint-authority account by `init`, so this no longer searches with `find_program_address`.

tokens/pda-mint-authority/pinocchio/program/src/instructions/mint.rs
MarkFeder: Done in a500411c — same change here: `mint` derives the PDA with `create_program_address` using the bump stored by `init`, no more `find_program_address`.

tokens/pda-mint-authority/pinocchio/tests/test.ts
MarkFeder: Done in a500411c — the Create instruction data is now built with `@solana/kit` codecs (`getStructEncoder` + `getU8Encoder`, and `addEncoderSizePrefix(getUtf8Encoder(), getU32Encoder())` for the borsh strings), and the `borsh` package is dropped — matching @amilz's approach in #675. Also moved the test stack to mocha + tsx and `@solana/kit` ^7 per AGENTS.md.

tokens/pda-mint-authority/pinocchio/tests/test.ts
MarkFeder: Done in a500411c — the ATA is now derived with `findAssociatedTokenPda` from `@solana-program/token` instead of hand-rolling the seeds. The metadata and master-edition PDAs stay on `getProgramDerivedAddress` (the kit built-in): there is no official `@solana-program` client for Metaplex Token Metadata (same reason its program id is hand-rolled here), and the umi-based `@metaplex-foundation/mpl-token-metadata` does not interop with the `@solana/kit` address type this test uses.

dev-jodee (approved):

amilz (approved):

PR #613, opened 2026-06-25
Add Pinocchio version of token-fundraiser example

boymak:
Adds a Pinocchio implementation of `tokens/token-fundraiser`, which previously only had an Anchor version — advancing the repo's goal of having every example available in every framework (Anchor, Pinocchio, Native).

## What's included
- `initialize`, `contribute`, `check_contributions` and `refund` instructions
- Mirrors the Anchor reference logic: 10% per-contributor cap, minimum target scaled by the mint's decimals, PDA-signed vault transfers, and account closing on settle / refund
- Follows the existing `escrow` / `transfer-tokens` Pinocchio conventions (manual borsh-compatible (de)serialization, bumps passed in the instruction data, `solana-bankrun` tests)
- Registers the program in the workspace `Cargo.toml` and links it from the root `README`

## Testing
- Program logic verified end-to-end on a local `solana-test-validator` (initialize → 2 contributions → over-cap rejection → settle-before-target rejection → refund with account close): **5/5 passing**.
- The `solana-bankrun` suite follows the same structure as the existing Pinocchio token examples.

## Heads-up for maintainers (not specific to this PR)
With the currently pinned `pinocchio 0.10.2` (whose deps require the Solana 4.0 build toolchain) plus `solana-bankrun 0.4.0`, running the bankrun suites locally traps with `unsupported BPF instruction` once a program actually executes. In CI the Pinocchio token suites currently report `0 passing` (the `describe(async …)` tests don't register before mocha finishes), so this stays hidden. Happy to help with a follow-up (e.g. migrating these suites to `litesvm`) if that'd be useful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5117,6 +5117,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "token-fundraiser-pinocchio-program"
+version = "0.1.0"
+dependencies = [
+ "pinocchio 0.10.2",
+ "pinocchio-associated-token-account",
+ "pinocchio-log",
+ "pinocchio-pubkey",
+ "pinocchio-system",
+ "pinocchio-token",
+]
+
+[[package]]
 name = "toml"
 version = "0.8.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ members = [
     # tokens
     "tokens/escrow/pinocchio/program",
     "tokens/transfer-tokens/pinocchio/program",
+    "tokens/token-fundraiser/pinocchio/program",
     "tokens/token-2022/mint-close-authority/native/program",
     "tokens/token-2022/non-transferable/native/program",
     "tokens/token-2022/default-account-state/native/program",
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ Allow two users to swap digital assets with each other, each getting 100% of wha
 
 Create a fundraiser account specifying a target mint and amount, allowing contributors to deposit tokens until the goal is reached.
 
-[anchor](./tokens/token-fundraiser/anchor)
+[anchor](./tokens/token-fundraiser/anchor) [pinocchio](./tokens/token-fundraiser/pinocchio)
 
 ### Minting a token from inside a program with a PDA as the mint authority
 
--- a/tokens/token-fundraiser/pinocchio/cicd.sh
+++ b/tokens/token-fundraiser/pinocchio/cicd.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# This script is for quick building & deploying of the program.
+# It also serves as a reference for the commands used for building & deploying Solana programs.
+# Run this bad boy with "bash cicd.sh" or "./cicd.sh"
+
+cargo build-sbf --manifest-path=./program/Cargo.toml --sbf-out-dir=./program/target/so
+solana program deploy ./program/target/so/program.so
--- a/tokens/token-fundraiser/pinocchio/package.json
+++ b/tokens/token-fundraiser/pinocchio/package.json
@@ -1,0 +1,25 @@
+{
+  "scripts": {
+    "test": "pnpm ts-mocha -p ./tsconfig.json -t 1000000 ./tests/test.ts",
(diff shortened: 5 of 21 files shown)


boymak:
Thanks for the review — both points were valid and are now fixed in e321f19:

1. **Vault substitution in `contribute`.** `contribute`, `check_contributions` and `refund` now verify the provided `vault` is a token account whose `owner` is the fundraiser PDA and whose `mint` is the raised mint (new `InvalidVault` error). This is the invariant the Anchor version gets from `associated_token::authority = fundraiser`. Previously a contributor could pass an account they control as the vault, keep their tokens, inflate `current_amount`, and later drain the real vault via a refund.

2. **`expectRevert` helper.** Rewrote it so it no longer catches its own sentinel — it now tracks whether the awaited promise rejected and throws if it didn't, so the negative-path tests actually assert.

Also added a regression test that a contribution with a substituted vault is rejected.

Verified end-to-end on a local `solana-test-validator` (7/7), including the vault-substitution rejection and a negative-control test confirming `expectRevert` is no longer vacuous.

boymak:
Good catch on the time gates — confirmed and fixed in c6e82bb.

Both `contribute` and `refund` had inverted comparisons (faithfully carried over from the Anchor reference):
- `contribute` now rejects once `elapsed_days >= duration` (campaign ended), so contributions are accepted **while the campaign is open**.
- `refund` now rejects while `elapsed_days < duration` (campaign still open), so refunds are allowed **only after it ends**.

As you noted, the bug was invisible because every test used `duration = 0` (where both inverted conditions happen to pass). The tests now use a non-zero, still-open campaign for the happy path and a second `duration = 0` (already-ended) campaign, which exercises **both** sides of each gate without needing to advance the clock.

Verified end-to-end on a local `solana-test-validator` (8/8): contributions accepted while open and rejected after the campaign ends, refunds rejected while open, plus the vault-ownership / per-contributor-cap / target guards.

Heads-up: the **Anchor and native** versions of `token-fundraiser` carry the same inverted conditions. Happy to open a follow-up PR fixing those for consistency if that's welcome.

boymak:
Added the rent-reclaim polish from the review in d812420:

- **`check_contributions` now closes the vault.** After draining the vault to the maker, it closes the (now-empty) vault token account in addition to the fundraiser account, returning both rents to the maker (mirrors the `escrow` example's `take_offer`). Previously the vault's rent was left locked.
- **Added a settlement happy-path test.** It funds a campaign to its target with 10 contributors (each at the 10% per-contributor cap), settles it, and asserts the maker receives the full amount and both the vault and the fundraiser account are closed.

Verified end-to-end on a local `solana-test-validator` (10 contributors → target reached → settle): the maker receives 30,000,000 and both the vault and fundraiser accounts are closed.

On the other note — closing the fundraiser account on `refund`: since the fundraiser is shared across all contributors and refunds are per-contributor, it can't be safely closed during an individual refund (other contributors may still need to reclaim their funds). This matches the Anchor reference, so I've left it as-is.

boymak:
Good catch — fixed in 05e4109.

You're right that an owner+mint check on the vault is insufficient: a maker could create a token account, assign its owner to the fundraiser PDA, pre-fund it to the target, settle against it, and close the fundraiser — stranding the real contributions.

Fix: the canonical vault address is now recorded in the fundraiser state at `initialize` (where the associated token program guarantees the provided vault is the canonical ATA for `(fundraiser, mint)`), and `contribute`, `check_contributions` and `refund` all verify the passed vault matches the recorded one. This replaces the previous owner+mint check with an exact-address check, so a substitute vault — canonical-looking or not — is rejected.

Verified on a local `solana-test-validator`:
- Settling against a fake fundraiser-owned vault pre-funded to the target is **rejected** (`InvalidVault`), and the real fundraiser stays open.
- A legitimate campaign funded to target by 10 contributors settles successfully and closes both the vault and the fundraiser.

Added a regression test (`Rejects settling against a non-canonical fundraiser-owned vault`) that reproduces the exact attack.

dev-jodee:
Thanks @boymak — we'd like to land this, but the repo changed substantially under it with #656 ([`b0d0c181`](https://github.com/solana-foundation/program-examples/commit/b0d0c181)). To get it mergeable, please rebase and conform to the new conventions (see AGENTS.md at the repo root): pinocchio 0.11 via the workspace deps, tests on mocha 11 + tsx (no ts-mocha/bankrun) using `@solana/kit` + litesvm 1.x — `tokens/create-token/pinocchio/tests/test.ts` is the reference pattern.

dev-jodee:
stale

PR #624, opened 2026-07-01
add pinocchio token-2022 mint-close-authority example

MarkFeder:
Adds a Pinocchio port of the `tokens/token-2022/mint-close-authority` example, alongside the existing `anchor` and `native` versions.

## What it does

A single instruction creates an SPL **Token-2022** mint carrying the `MintCloseAuthority` extension, so a designated authority can later close the mint and reclaim its rent.

## Notes

Token-2022 has no Pinocchio wrapper crate (`pinocchio-token` targets the legacy SPL Token program), so the Token-2022 instructions are built by hand and CPI'd:

1. `CreateAccount` for the mint, sized to `202` bytes (base `Account` length 165 + account-type byte + one `MintCloseAuthority` TLV entry) and owned by the Token-2022 program.
2. `InitializeMintCloseAuthority` (variant `25`) — must run **before** the mint is initialized.
3. `InitializeMint` (variant `0`).

The bankrun test asserts the resulting mint is owned by Token-2022, is exactly 202 bytes, has the correct decimals, and that the extension header (account type `Mint`, TLV type `MintCloseAuthority`) and stored close authority were written correctly.

### DIFF
--- a/tokens/token-2022/mint-close-authority/pinocchio/program/src/instructions/create_mint.rs
+++ b/tokens/token-2022/mint-close-authority/pinocchio/program/src/instructions/create_mint.rs
@@ -1,0 +1,118 @@
+use alloc::vec::Vec;
+
+use pinocchio::{
+    cpi::invoke,
+    error::ProgramError,
+    instruction::{InstructionAccount, InstructionView},
+    sysvars::{rent::Rent, Sysvar},
+    AccountView, ProgramResult,
+};
+use pinocchio_log::log;
+use pinocchio_system::instructions::CreateAccount;
+
+use crate::instructions::{CreateTokenArgs, MINT_SIZE, TOKEN_2022_PROGRAM_ID};
+
+/// Token-2022 instruction discriminators (variants of the program's instruction
+/// enum) that this example builds by hand.
+const INITIALIZE_MINT: u8 = 0;
+const INITIALIZE_MINT_CLOSE_AUTHORITY: u8 = 25;
+
+/// Creates a new SPL Token-2022 mint that carries the `MintCloseAuthority`
+/// extension, allowing the configured authority to later close the mint account
+/// and reclaim its rent.
+///
+/// Accounts:
+///   0. `[signer, writable]` mint account (a fresh keypair to initialize)
+///   1. `[]`                 mint authority (also set as the freeze authority)
+///   2. `[]`                 close authority (allowed to close the mint later)
+///   3. `[signer, writable]` payer (funds the new account)
+///   4. `[]`                 rent sysvar
+///   5. `[]`                 system program
+///   6. `[]`                 Token-2022 program
+///
+/// Instruction data: Borsh `[decimals: u8]`.
+pub fn create_mint(accounts: &[AccountView], data: &[u8]) -> ProgramResult {
+    // `system_program` and `token_program` are unused directly, but must be
+    // supplied so they are present in the transaction for the CPIs below.
+    let [mint_account, mint_authority, close_authority, payer, rent_sysvar, _system_program, _token_program] =
+        accounts
+    else {
+        return Err(ProgramError::NotEnoughAccountKeys);
+    };
+
+    let args = CreateTokenArgs::parse(data)?;
+
+    // Fund the mint account with enough lamports to stay rent-exempt at the
+    // extended size, and create it owned by the Token-2022 program.
+    let lamports = Rent::get()?.try_minimum_balance(MINT_SIZE)?;
+
+    log!("Creating mint account");
+    CreateAccount {
+        from: payer,
+        to: mint_account,
+        lamports,
+        space: MINT_SIZE as u64,
+        owner: &TOKEN_2022_PROGRAM_ID,
+    }
+    .invoke()?;
+
+    // The `MintCloseAuthority` extension must be initialized *before* the mint
+    // itself — extensions live in the space past the base mint and Token-2022
+    // rejects initializing them once `InitializeMint` has run.
+    log!("Initializing mint close authority extension");
+    let close_authority_data = build_initialize_close_authority_data(close_authority);
+    let close_authority_accounts = [InstructionAccount::writable(mint_account.address())];
+    invoke(
+        &InstructionView {
+            program_id: &TOKEN_2022_PROGRAM_ID,
+            accounts: &close_authority_accounts,
+            data: &close_authority_data,
+        },
+        &[mint_account],
+    )?;
+
+    log!("Initializing mint");
+    let mint_data = build_initialize_mint_data(mint_authority, args.decimals);
+    let mint_accounts = [
+        InstructionAccount::writable(mint_account.address()),
+        InstructionAccount::readonly(rent_sysvar.address()),
+    ];
+    invoke(
+        &InstructionView {
+            program_id: &TOKEN_2022_PROGRAM_ID,
+            accounts: &mint_accounts,
+            data: &mint_data,
+        },
+        &[mint_account, rent_sysvar],
+    )?;
+
+    log!("Mint created");
+    Ok(())
+}
+
+/// Serializes an `InitializeMintCloseAuthority` instruction (variant 25).
+///
+/// Layout: `[25] close_authority: COption<Pubkey>`, where a present pubkey is
+/// encoded as a `1` tag byte followed by the 32-byte key.
+fn build_initialize_close_authority_data(close_authority: &AccountView) -> Vec<u8> {
+    let mut data = Vec::with_capacity(34);
+    data.push(INITIALIZE_MINT_CLOSE_AUTHORITY);
+    data.push(1); // COption::Some
+    data.extend_from_slice(close_authority.address().as_ref());
+    data
+}
+
+/// Serializes an `InitializeMint` instruction (variant 0).
+///
+/// Layout: `[0] decimals: u8 mint_authority: Pubkey freeze_authority:
+/// COption<Pubkey>`. The mint authority doubles as the freeze authority, matching
+/// the `native` example.
+fn build_initialize_mint_data(mint_authority: &AccountView, decimals: u8) -> Vec<u8> {
+    let mut data = Vec::with_capacity(67);
+    data.push(INITIALIZE_MINT);
+    data.push(decimals);
+    data.extend_from_slice(mint_authority.address().as_ref());
+    data.push(1); // freeze_authority: COption::Some
+    data.extend_from_slice(mint_authority.address().as_ref());
+    data
+}
--- a/tokens/token-2022/mint-close-authority/pinocchio/program/src/instructions/create_mint.rs
+++ b/tokens/token-2022/mint-close-authority/pinocchio/program/src/instructions/create_mint.rs
@@ -4,7 +4,7 @@ use pinocchio::{
     cpi::invoke,
     error::ProgramError,
     instruction::{InstructionAccount, InstructionView},
-    sysvars::{rent::Rent, Sysvar},
+    sysvars::rent::{ACCOUNT_STORAGE_OVERHEAD, DEFAULT_LAMPORTS_PER_BYTE},
     AccountView, ProgramResult,
 };
 use pinocchio_log::log;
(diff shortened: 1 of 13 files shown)


tokens/token-2022/mint-close-authority/pinocchio/tests/test.ts
```diff
@@ -0,0 +1,91 @@
...
+    // The extension header marks this as a Mint carrying MintCloseAuthority.
+    assert.equal(mintData[ACCOUNT_TYPE_OFFSET], ACCOUNT_TYPE_MINT);
+    assert.equal(mintData.readUInt16LE(TLV_TYPE_OFFSET), MINT_CLOSE_AUTHORITY_EXTENSION);
+
+    // The configured close authority was stored in the extension.
+    const storedCloseAuthority = mintData.subarray(TLV_VALUE_OFFSET, TLV_VALUE_OFFSET + 32);
+    assert.deepEqual(new Uint8Array(storedCloseAuthority), payer.publicKey.toBytes());
+
+    console.log("Mint address:", mintKeypair.publicKey.toBase58());
+  });
```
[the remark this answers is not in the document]
MarkFeder: Good catch — fixed in 1b2f802. The test now generates a distinct `closeAuthority` keypair for account index 2, so the stored-authority assertion genuinely verifies the close authority is sourced from index 2 rather than the mint authority/payer (which remain `payer`).

tokens/token-2022/mint-close-authority/pinocchio/tests/test.ts
```diff
@@ -0,0 +1,91 @@
...
+};
+
+// Token-2022 lays a mint with one extension out as:
+//   base account length (165) + account-type byte (1) + TLV entry (36) = 202
+const EXTENDED_MINT_SIZE = 202;
+const ACCOUNT_TYPE_OFFSET = 165; // 1 == Mint
+const TLV_TYPE_OFFSET = 166; // u16 LE, 3 == MintCloseAuthority
+const TLV_VALUE_OFFSET = 170; // 32-byte close authority pubkey
+const DECIMALS_OFFSET = 44; // in the base mint layout
+const MINT_CLOSE_AUTHORITY_EXTENSION = 3;
```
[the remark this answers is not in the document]
MarkFeder: Agreed — added `assert.equal(mintData.readUInt16LE(168), 32)` (TLV_LENGTH_OFFSET) in 1b2f802 so the 2-byte TLV length header is exercised explicitly.

tokens/token-2022/mint-close-authority/pinocchio/program/src/instructions/create_mint.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+/// Instruction data: Borsh `[decimals: u8]`.
+pub fn create_mint(accounts: &[AccountView], data: &[u8]) -> ProgramResult {
+    // `system_program` and `token_program` are unused directly, but must be
+    // supplied so they are present in the transaction for the CPIs below.
+    let [mint_account, mint_authority, close_authority, payer, rent_sysvar, _system_program, _token_program] =
+        accounts
+    else {
+        return Err(ProgramError::NotEnoughAccountKeys);
+    };
```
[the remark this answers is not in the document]
MarkFeder: The premise here isn't quite accurate: `_token_program` (index 6) is **not** forwarded to the CPIs as the callee. Each `InstructionView` sets `program_id: &TOKEN_2022_PROGRAM_ID` — the hard-coded constant — and the account isn't even passed in the `invoke` infos slices (`&[mint_account]` / `&[mint_account, rent_sysvar]`). So a spoofed account at index 6 has no effect; the CPIs always target the real Token-2022 program by construction. Index 6 is present only to mirror the native example's account layout. No equality check is needed here since dispatch is already pinned to the constant.

MarkFeder:
Hi @Perelyn-sama — thanks again for merging my previous pinocchio examples! This one adds the `token-2022/mint-close-authority` pinocchio port (the first Token-2022 example in the pinocchio series). All checks are green except a single red X on the ASM `build-and-test-group-0` — that's the pre-existing repo-wide ASM failure, unrelated to this diff, which #625 fixes. Greptile has it at 5/5 "safe to merge." Would appreciate a review whenever you get a chance. Thanks!

MarkFeder:

tokens/token-2022/mint-close-authority/pinocchio/tests/test.ts
```diff
@@ -0,0 +1,102 @@
+import { Buffer } from "node:buffer";
+import {
+  Keypair,
+  PublicKey,
+  SYSVAR_RENT_PUBKEY,
+  SystemProgram,
+  Transaction,
+  TransactionInstruction,
+} from "@solana/web3.js";
```
dev-jodee: Hey Mark, thanks for the work, realized (a bit too late for the already merged project) that you're using web3js, this isn't the proper way to do it anymore, I'd encourage you to look at kit (https://github.com/anza-xyz/kit) and if you could update all the other examples as well, make sure we use the latest standards ! (Would be great if you could do it for the already merged pr from yesterday as well)

dev-jodee (changes requested):

tokens/token-2022/mint-close-authority/pinocchio/tests/test.ts
MarkFeder: Thanks @dev-jodee — done for this one (`59d9542`). The test now builds everything with `@solana/kit`: addresses, the instruction, the transaction message, and signing (`generateKeyPairSigner`/`createKeyPairSignerFromBytes`, `createTransactionMessage` + `pipe`, `signTransactionMessageWithSigners`, `getTransactionEncoder`). Bumped the example's `tsconfig` target/lib to es2020 for kit's BigInt usage; still 1 passing under bankrun.

One thing worth flagging before I roll this out to the rest: `solana-bankrun@0.3.1`'s `BanksClient` is still typed against web3.js — `start()` and `getAccount()` take a `PublicKey`, and `processTransaction()` takes a `VersionedTransaction` — and it exposes no byte-level entrypoint. So a thin, unavoidable web3.js interop shim remains *around bankrun itself*: kit builds and signs the tx, then I hand its wire bytes to bankrun via `VersionedTransaction.deserialize`. Everything the example teaches is kit; web3.js only survives as the harness adapter.

Happy to apply this same pattern across the other pinocchio examples + the merged create-token. If you'd rather also move these off bankrun (e.g. to litesvm) so web3.js drops entirely, say the word and I'll factor that in instead.

dev-jodee: Yeah, honestly we should have the examples with `kit` and `litesvm` for testing framework, I think with both of those we'll have a good template

Really appreciate the work :) 

You can do the change on this one, once we confirm + merge you can apply to the rest (just in case something else comes up, lets make sure this one is perfect and then you can move to the other ones )

MarkFeder: Done — migrated this one fully to `litesvm` in `cd5cf69`, so `@solana/web3.js` is now gone entirely (both from the code and from `package.json`).

The nice part is that `litesvm@1.3` is kit-native, so the two compose with no shim:
- the signed kit transaction from `signTransactionMessageWithSigners` goes **straight** into `svm.sendTransaction(...)` — no more `VersionedTransaction.deserialize` round-trip;
- `svm.getAccount()` / `svm.airdrop()` take kit `Address`es and return kit `EncodedAccount`s;
- `svm.setTransactionMessageLifetimeUsingLatestBlockhash(...)` slots right into the `pipe(...)`.
- `new LiteSVM()`'s standard runtime already bundles Token-2022, so there's no fixture to load beyond the program `.so`.

A couple of supporting tweaks so it's a clean template to copy from:
- pinned `@solana/kit` to `^6.10.0` (litesvm's kit major) so there's a single kit in the tree — no dual-version type/shape mismatches;
- bumped `typescript` to `^5` and `lib` to `es2022`+`dom` (kit's types need both) and added `@types/node`; the suite is now type-clean under `tsc --noEmit`, not just transpile-only.

Verified locally: `cargo build-sbf` + `ts-mocha` → **1 passing** (program executes both CPIs, mint layout/close-authority assertions hold), plus `tsc --noEmit` and `biome` clean.

Once you're happy with this one and it's merged, I'll roll the same kit + litesvm pattern out across the rest. 🙏

dev-jodee (changes requested):
need signed commits please

tokens/token-2022/mint-close-authority/pinocchio/tests/test.ts
MarkFeder: Signed now 🙏 — all five commits on the branch are re-signed and show **Verified** on GitHub (the two oldest, `d98603d`/`b40292f`, were unsigned; I force-rebased onto the same base to re-sign them, tree is byte-identical so no content changed). The kit + litesvm migration is unchanged. Ready for another look whenever you have a moment.

tokens/token-2022/mint-close-authority/pinocchio/program/src/instructions/create_mint.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let args = CreateTokenArgs::parse(data)?;
+
+    // Fund the mint account with enough lamports to stay rent-exempt at the
+    // extended size, and create it owned by the Token-2022 program. Rent is
+    // computed with integer math using the network's default parameters
+    // (`DEFAULT_LAMPORTS_PER_BYTE` already folds in the 2-year exemption
+    // threshold). We avoid `Rent::try_minimum_balance`, whose floating-point
+    // exemption-threshold path emits an instruction the bankrun test VM rejects
```
dev-jodee: since we're running litesvm tests instead of bankrun now, I think you can use the Rent:: function

dev-jodee: something like 

```
use pinocchio::sysvars::{rent::Rent, Sysvar};

  let rent = Rent::get()?;
  let lamports = rent.try_minimum_balance(Mint::LEN)?;
```

tokens/token-2022/mint-close-authority/pinocchio/tests/test.ts
```diff
@@ -0,0 +1,129 @@
...
+  setTransactionMessageFeePayerSigner,
+  signTransactionMessageWithSigners,
+} from "@solana/kit";
+import * as borsh from "borsh";
+import { assert } from "chai";
+import { FailedTransactionMetadata, LiteSVM } from "litesvm";
+
+// LiteSVM's standard runtime bundles the SPL programs, so Token-2022 is already
+// loaded — its ID is hard-coded here to avoid pulling in @solana/spl-token.
+const TOKEN_2022_PROGRAM_ID = address("TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb");
```
dev-jodee: I think a lot of your constants you should just use the packages that already have those, I know it add dependencies, but to keep it as the "best" possible template, we should showcase all available official packages too.

I.e. @solana-program/token-2022, @solana/sysvars, @solana-program/system

and from kit stuff like ```getMintDecoder().decode(mintData)``` to get the offsets, etc. cc @amilz is the professional `kit` chief, so he prob has some feedback too

dev-jodee (changes requested):
couple more things :) we're getting close ! (also apply them to all the other prs too when it makes sense)

tokens/token-2022/mint-close-authority/pinocchio/program/src/instructions/create_mint.rs
MarkFeder: Done in 807774e — swapped the integer-math workaround for exactly this:

```rust
use pinocchio::sysvars::{rent::Rent, Sysvar};

let rent = Rent::get()?;
let lamports = rent.try_minimum_balance(MINT_SIZE)?;
```

The integer math was only there to dodge the f64 opcode the old bankrun VM rejected; litesvm runs the real syscall fine.

tokens/token-2022/mint-close-authority/pinocchio/tests/test.ts
MarkFeder: Done in 807774e. The test now pulls ids/decoders from the official packages instead of hand-rolled constants:

- `TOKEN_2022_PROGRAM_ADDRESS` from `@solana-program/token-2022`
- `SYSTEM_PROGRAM_ADDRESS` from `@solana-program/system`
- `SYSVAR_RENT_ADDRESS` from `@solana/sysvars`
- `getMintDecoder().decode(mintAccount.data)` to read `decimals` and the `MintCloseAuthority` extension (`unwrapOption(mintState.extensions)`), dropping all the raw byte-offset constants.

One version note for reviewers: `litesvm@1.3` pins `@solana/kit@^6.10`, so I pinned `@solana-program/token-2022@^0.12` / `@solana-program/system@^0.12` / `@solana/sysvars@^6.10` (the `0.13`/`7.x` lines peer-require kit 7 and would split the tree). Also bumped tsconfig `moduleResolution` to `bundler` so the packages subpath exports type-check. `tsc --noEmit`, biome, and `ts-mocha` (1 passing) all green. cc @amilz in case you have `kit` idiom feedback.

dev-jodee (approved):

dev-jodee:
thanks for the work @MarkFeder, can you do the pr to fix the previous one we already merged with the feedback from this one please :)

MarkFeder:
Thanks for merging @dev-jodee! 🙏 Opened the follow-up you asked for — #641 brings the already-merged create-token example up to this same `kit` + `litesvm` + official-package template (real `Rent::get()`, `@solana-program/token`/`@solana-program/system`, `getProgramDerivedAddress`, tsconfig `moduleResolution: bundler`). Verified locally: `cargo build-sbf` + `ts-mocha` → 2 passing, `tsc`/biome/fmt/clippy clean, frozen-lockfile OK, commit signed.

PR #657, opened 2026-07-29
chore: upgrade nft-meta-data-pointer app to Next.js 16 and React 18.3

Harsh-H-Shah:
## Summary

This Token-2022 example's frontend was on Next.js 13.4.4 / React 18.2.0 (mid-2023) — the only browser app in this repo, and the one place where dependency staleness is a security question (wallet-connected) rather than pure hygiene. This PR brings it to Next.js 16.2.12 / React 18.3.1.

Stays on React 18.x rather than 19: `@magicblock-labs/gum-react-sdk` (used for the session-key feature) hard-pins `react: ^18.2.0` in every published release up to latest (`3.0.10`), with no React 19 support. Next.js itself doesn't force the issue — both Next 15 and 16 still accept `react: ^18.2.0`.

## What's in scope vs. what surfaced

The version jump itself is small. Getting `next build` to pass under it was not — Next 16's stricter type-checking and Turbopack bundling caught several **pre-existing** bugs that had never been exercised before. I've tried to keep those fixes minimal and mechanical rather than rewriting anything.

### Direct upgrade
- `next` 13.4.4 → 16.2.12, `react`/`react-dom` 18.2.0 → 18.3.1
- `eslint` 8.41.0 → 9.39.5, `eslint-config-next` → 16.2.12
- `.eslintrc.json` → `eslint.config.mjs` (flat config) — `next lint` was removed entirely in Next 16
- `typescript` 5.0.4 → 5.9.3 (needed — several `@solana/*` transitive deps require `typescript >= 5.4.0`)

### Pre-existing bugs this surfaced
- **`idl/extension_nft.ts`** hand-writes this program's Anchor IDL and was in an old schema (`isOptional` instead of `optional`, account structs embedded directly instead of pointing into a `types` array, `"publicKey"` instead of `"pubkey"` as a type literal, and missing the now-required `address`/`metadata`/`discriminator` fields) — incompatible with the already-installed `@anchor-lang/core` client, which silently collapsed every account's decoded type to `never`. Restructured to the current schema. **Instruction/account discriminators are computed via `sha256("global:"|"account:" + name).slice(0, 8)`, matching Anchor's own algorithm exactly — not guessed** — worth double-checking against the real deployed program's IDL if you have it handy.
- **`utils/anchor.ts`**: the `Program` constructor no longer takes `programId` as a positional argument now that `idl.address` is required and used instead.
- **`@coral-xyz/anchor` and `axios`** were imported in a couple of files but never declared as dependencies — they only resolved before by accident, via `node_modules` hoisting under Next 13's looser module resolution. Now declared properly (`axios`) or replaced with the already-used `@anchor-lang/core`/`@solana/web3.js` equivalents.
- **`@solana/wallet-adapter-wallets`** (the mega-package bundling every hardware/mobile wallet adapter) pulls in a Ledger dependency with a broken ESM export under Turbopack, crashing the build. This app only ever uses Phantom + Solflare, so swapped to the standalone `@solana/wallet-adapter-phantom` / `@solana/wallet-adapter-solflare` packages instead of fighting the bundler over unused code.
- **`gum-react-sdk`'s `createSession()`** signature drifted between the pinned `^3.0.4` range and the resolved `3.0.10`: the 2nd parameter changed from a `boolean` to `topUpLamports: number`.
- **`ChopTreeButton.tsx`**: one `chopTree` instruction call was missing the required `systemProgram`/`tokenProgram` accounts (would have failed on-chain), and an NFT-authority field (typed as `string | { address: string }`) was read unsafely in two places without checking which shape it was.

## Requesting extra scrutiny on

- The discriminator values and restructured `idl/extension_nft.ts` — I computed these correctly against Anchor's known algorithm, but I don't have a way to test them against the actual deployed program (`H31ofLpWqeAzF2Pg54HSPQGYifJad843tTJg8vCYVoh3`) in this environment.
- The `ChopTreeButton.tsx` account/null-safety fixes — logically sound based on the sibling code path in the same file, but not exercised against a live wallet.

## Test plan

- [x] `next build` passes: full TypeScript check + static generation of all routes
- [x] `next dev` serves the homepage (`GET / 200`), and the SSR output contains real Chakra UI + wallet-adapter markup — the full provider tree mounts and renders without throwing
- [x] `eslint .` passes with 0 errors (3 pre-existing `exhaustive-deps` warnings left as-is, unrelated to this change)
- [ ] **Live browser click-through / wallet connect** — not possible in the environment this was built in (no Chromium available). Needs a manual smoke test before merging.

### DIFF
--- a/tokens/token-2022/nft-meta-data-pointer/app/components/SessionKeyButton.tsx
+++ b/tokens/token-2022/nft-meta-data-pointer/app/components/SessionKeyButton.tsx
@@ -1,6 +1,7 @@
 import { Button } from '@chakra-ui/react';
 import { useSessionWallet } from '@magicblock-labs/gum-react-sdk';
 import { useWallet } from '@solana/wallet-adapter-react';
+import { LAMPORTS_PER_SOL } from '@solana/web3.js';
 import { useState } from 'react';
 import { useGameState } from '@/contexts/GameStateProvider';
 import { program } from '@/utils/anchor';
@@ -13,11 +14,11 @@ const SessionKeyButton = () => {
 
     const handleCreateSession = async () => {
         setIsLoading(true);
-        const topUp = true;
+        const topUpLamports = 0.01 * LAMPORTS_PER_SOL;
         const expiryInMinutes = 600;
 
         try {
-            const session = await sessionWallet.createSession(program.programId, topUp, expiryInMinutes);
+            const session = await sessionWallet.createSession(program.programId, topUpLamports, expiryInMinutes);
             console.log('Session created:', session);
         } catch (error) {
             console.error('Failed to create session:', error);
--- a/tokens/token-2022/nft-meta-data-pointer/app/.eslintrc.json
+++ b/tokens/token-2022/nft-meta-data-pointer/app/.eslintrc.json
@@ -1,3 +1,0 @@
-{
-    "extends": "next/core-web-vitals"
-}
--- a/tokens/token-2022/nft-meta-data-pointer/app/components/ChopTreeButton.tsx
+++ b/tokens/token-2022/nft-meta-data-pointer/app/components/ChopTreeButton.tsx
@@ -2,7 +2,7 @@ import { Button, HStack, VStack } from '@chakra-ui/react';
 import { useSessionWallet } from '@magicblock-labs/gum-react-sdk';
 import { TOKEN_2022_PROGRAM_ID } from '@solana/spl-token';
 import { useConnection, useWallet } from '@solana/wallet-adapter-react';
-import { PublicKey } from '@solana/web3.js';
+import { PublicKey, SystemProgram } from '@solana/web3.js';
 import Image from 'next/image';
 import { useCallback, useState } from 'react';
 import { useGameState } from '@/contexts/GameStateProvider';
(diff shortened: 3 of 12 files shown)


tokens/token-2022/nft-meta-data-pointer/app/components/SessionKeyButton.tsx
```diff
@@ -1,6 +1,7 @@
 import { Button } from '@chakra-ui/react';
 import { useSessionWallet } from '@magicblock-labs/gum-react-sdk';
 import { useWallet } from '@solana/wallet-adapter-react';
+import { LAMPORTS_PER_SOL } from '@solana/web3.js';
```
dev-jodee, line 4: we should remove web 3js dependency, move to kit

dev-jodee (changes requested):
can you remove the web3js dependency and move to kit

tokens/token-2022/nft-meta-data-pointer/app/components/SessionKeyButton.tsx
Harsh-H-Shah: Happy to take this on, but wanted to flag the actual scope before diving in, since it's bigger than a dependency swap.

`@anchor-lang/core`'s `Program`/`Provider` API is hard-typed to `@solana/web3.js` — `Provider.connection` is typed as web3.js's `Connection`, `publicKey` as `PublicKey`, and every method around `Transaction | VersionedTransaction`. Every `program.methods.chopTree(...)` call in this app goes through that class, so there's no way to drop web3.js while keeping `@anchor-lang/core` — they're coupled at the type level. `@solana/wallet-adapter-react`'s `useWallet()` is the same story (`PublicKey`, `Connection`-typed `sendTransaction`).

So "move to kit" here really means:
- Dropping `@anchor-lang/core` for a Codama-generated `@solana/kit`-native client from this program's IDL (I see `games/world-cup` already does this in-repo, happy to follow that pattern)
- Rewriting every instruction call site to kit's functional builders
- Replacing `wallet-adapter-react` with something kit-native (wallet-standard-first / ConnectorKit)
- Rewriting `WrappedConnection` (currently extends web3.js's `Connection`) for the DAS calls

That's a rebuild of this example's client architecture, not something I think should ride along with the Next.js/React upgrade in this PR — it'd make this much harder to review and much easier to regress something. Want it as a separate follow-up PR once this one's in? Happy to scope that out properly (probably worth its own design pass given the wallet-adapter replacement question) rather than rushing it in here.

dev-jodee: ah fair, for now it's fine, since anchor has a hard dep, thanks for the reply

dev-jodee (approved):

PR #661, opened 2026-08-02
tokens: add merkle-tree-token-claimer anchor example

brimigs:
## Problem

The Cosmos app-chain migration guide under review in solana-foundation/solana-com#1251 links a working Merkle token claimer as its reference implementation, but that example currently lives in a personal repo (`brimigs/cosmos-migration-guide`). Review feedback on that PR asked for the example to move into program-examples so the guide can point at an officially maintained home.

## Summary of changes

Adds `tokens/merkle-tree-token-claimer` (anchor): distribute a snapshot of token balances from one funded vault and a single 32-byte Merkle root, the claim pattern behind large airdrops and chain migrations.

**Program** (three instructions):
- `initialize_airdrop_data` — stores the Merkle root, mints the full claimable supply into a vault ATA owned by the program PDA, then revokes the mint authority so supply is fixed at launch.
- `update_tree` — replaces the root, permitted only before the first claim so live proofs never invalidate.
- `claim_airdrop` — recomputes the leaf from `signer + amount`, verifies the sha256 proof against the stored root, transfers via PDA-signed `transfer_checked`, and writes a per-index `claim_receipt` PDA that blocks double-claims. Claims are additionally capped by the initialized total.

**Tests** (`tests/litesvm.test.ts`, mocha + tsx + anchor-litesvm per repo conventions): initialization state and the mint lock, pre-claim root updates, payouts recording receipts across two claimants, duplicate-claim rejection, stolen-proof rejection (attacker replaying someone else's proof), and the post-claim root freeze. All assertions check real post-state (account fields, token balances, vault deltas).

**Tooling**: `scripts/generate-merkle-tree.ts` turns a snapshot JSON into the on-chain root plus a proof per claimant (`pnpm generate-tree scripts/sample-snapshot.json out.json`), sharing the same tree implementation the tests use (`tests/merkle.ts`), which matches the on-chain verifier byte for byte.

Also registers the program crate in `.github/.workspace-ignore` (nested anchor workspace, same as the other tokens examples) and adds the example to the root README.

## Verification

- `pnpm install` (lockfile committed), `anchor build --ignore-keys` ✅
- `pnpm exec tsc --noEmit -p tsconfig.json` ✅
- LiteSVM suite: 4 passing ✅
- `cargo fmt --check` clean, prettier (root config) clean

### DIFF
--- a/tokens/merkle-tree-token-claimer/anchor/scripts/generate-merkle-tree.ts
+++ b/tokens/merkle-tree-token-claimer/anchor/scripts/generate-merkle-tree.ts
@@ -1,0 +1,113 @@
+/**
+ * Merkle tree generator for token claims.
+ *
+ * Reads a snapshot JSON file and produces:
+ * 1. The Merkle root (stored on-chain by initialize_airdrop_data)
+ * 2. An individual proof for each address (served to users by your claim UI)
+ *
+ * Usage: pnpm generate-tree <snapshot.json> <output.json>
+ * Example: pnpm generate-tree scripts/sample-snapshot.json merkle-output.json
+ */
+
+import * as fs from 'node:fs';
+import { PublicKey } from '@solana/web3.js';
+import { leafBytes, MerkleTree } from '../tests/merkle.ts';
+
+interface SnapshotEntry {
+    source_address: string;
+    solana_address: string;
+    amount: string | number;
+}
+
+interface Snapshot {
+    snapshot_height: number;
+    chain_id: string;
+    timestamp: string;
+    entries: SnapshotEntry[];
+}
+
+interface ProofEntry {
+    solana_address: string;
+    source_address: string;
+    amount: string;
+    index: number;
+    proof: string;
+}
+
+const U64_MAX = 2n ** 64n - 1n;
+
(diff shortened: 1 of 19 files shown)


tokens/merkle-tree-token-claimer/anchor/scripts/generate-merkle-tree.ts
```diff
@@ -0,0 +1,117 @@
...
+ * Reads a snapshot JSON file and produces:
+ * 1. The Merkle root (stored on-chain by initialize_airdrop_data)
+ * 2. An individual proof for each address (served to users by your claim UI)
+ *
+ * Usage: pnpm generate-tree <snapshot.json> <output.json>
+ * Example: pnpm generate-tree scripts/sample-snapshot.json merkle-output.json
+ */
+
+import * as fs from 'node:fs';
+import { PublicKey } from '@solana/web3.js';
```
dev-jodee: can we move away from kit ?

dev-jodee (changes requested):
can we move web3js to kit so it fits with the rest of the repo

dev-jodee (approved):

PR #664, opened 2026-08-03
feat(gacha): add provably-fair gacha (pack-pull) example

amilz:
Adds a complete provably-fair **gacha** (loot-box / pack-pull) game example — the on-chain mechanic behind Solana RWA pack platforms like Collector Crypt and Phygitals. An admin configures a pool of fixed-weight reward tiers and a fixed entry fee; buyers open pulls revealed via RFC 9381 ECVRF anchored in the deployed [`cc-vrf`](https://vrf.collectorcrypt.com) registry by CPI, and prizes are minted as Token-2022 NFTs carrying a `rarity` metadata field.

## What's included

- **Pinocchio program** — full commit-reveal lifecycle (`init_pool`, `buy_pull`, `settle_pull`, `refund_pull`, `claim_prize`, `withdraw_fees`) with Anchor-compatible self-CPI events and a hand-built CPI into cc-vrf's `commit_proof_with_beta`.
- **Codama pipeline + clients** — IDL-driven TypeScript and Rust client generation, plus a `@solana/gacha` kit plugin used by both scripts and the webapp.
- **Test coverage** — LiteSVM integration suite (lifecycle + negatives) and a Light-stack suite exercising the real `settle_pull` → cc-vrf → Light CPI chain with genuine validity proofs; Rust↔TS parity for `select_tier`/`derive_alpha` pinned by shared fixtures.
- **Next.js webapp** — buy/reveal/verify UI with a server-side pull orchestrator (submit → settle → mint), SIMD All-Stars collectible card reveal, and an admin panel for pool creation and fee withdrawal.
- **Docs** — trust-model writeup (detection-not-prevention randomness model), READMEs, and CI wiring.

## Trust model

The program accepts the registered operator's signed `beta`, while buyer entropy in `alpha`, fixed weights, one-reveal-per-pull enforcement via cc-vrf/Light, and a refund escape hatch close the gaps detection alone leaves open. Verification is fully reproducible off-chain from emitted events.

Closes DEV-776


games/gacha/pinocchio/tests/integration-tests/src/utils/constants.rs
```diff
@@ -0,0 +1,18 @@
+use solana_pubkey::Pubkey;
+
+pub static PROGRAM_ID: Pubkey = Pubkey::new_from_array(crate::ID.to_bytes());
+pub static SYSTEM_PROGRAM_ID: Pubkey = Pubkey::new_from_array([0u8; 32]);
```
dev-jodee: Instead of declaring as constants, we should prob import the packages and use the constants in there. Cleaner imo

games/gacha/pinocchio/tests/integration-tests/src/utils/pda.rs
```diff
@@ -0,0 +1,27 @@
...
+
+pub fn get_pool_pda(admin: &Pubkey) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&[POOL_SEED, admin.as_ref()], &PROGRAM_ID)
+}
+
+pub fn get_vault_pda(admin: &Pubkey) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&[VAULT_SEED, admin.as_ref()], &PROGRAM_ID)
+}
+
+pub fn get_pull_pda(pool: &Pubkey, buyer: &Pubkey, index: u64) -> (Pubkey, u8) {
```
dev-jodee: I might be wrong, but I think codama exports the pda derivation utils?

games/gacha/pinocchio/SECURITY.md
```diff
@@ -0,0 +1,9 @@
+# Security Policy
+
+## Reporting security problems
+
+**DO NOT CREATE A GITHUB ISSUE** to report a security problem.
```
dev-jodee: Don't think we want a security.md per program examples, this is at the repo root level

games/gacha/pinocchio/LICENSE
```diff
@@ -0,0 +1,9 @@
+MIT License
```
dev-jodee: LICENSE is at the top of the repo, don't think we need per program

games/gacha/pinocchio/tests/integration-tests/src/utils/constants.rs
```diff
@@ -0,0 +1,18 @@
+use solana_pubkey::Pubkey;
```
dev-jodee: Should we be using Address instead of pubkey?

games/gacha/pinocchio/tests/integration-tests/src/utils/idl.rs
```diff
@@ -0,0 +1,34 @@
...
+
+pub struct IdlAccount {
+    pub index: usize,
+    pub name: String,
+    pub is_writable: bool,
+    pub is_signer: bool,
+}
+
+/// The ordered account metas the IDL declares for an instruction.
+pub fn instruction_accounts(instruction: &str) -> Vec<IdlAccount> {
```
dev-jodee, line 20: what's this use for ?

games/gacha/pinocchio/program/src/state/common.rs
```diff
@@ -0,0 +1,88 @@
...
+pub const MINT_SEED: &[u8] = b"mint";
+
+/// One-byte discriminator identifying the type of a program-owned account.
+///
+/// Stored at byte offset 0 of every data-carrying account created by this program.
+#[repr(u8)]
+#[derive(Clone, Copy, PartialEq, Debug, CodamaType)]
+pub enum AccountDiscriminator {
+    /// [`Pool`](super::pool::Pool) account.
+    Pool = 0,
```
dev-jodee, line 24: usually you start account discriminator > 0, because if you close the pda and zero out the account, then if you check data[0] == 0, it would pass as if its a pool, even tho its a zeroed out account

games/gacha/pinocchio/program/src/state/pool.rs
```diff
@@ -0,0 +1,139 @@
...
+#[codama(seed(name = "admin", type = public_key))]
+pub struct Pool {
+    /// Account type discriminator ([`AccountDiscriminator::Pool`]).
+    pub discriminator: u8,
+    /// PDA bump seed.
+    pub bump: u8,
+    /// Number of active tiers (leading entries of `weights`).
+    pub tier_count: u8,
+    /// Admin that configured and owns this pool.
+    pub admin: Address,
```
dev-jodee, line 30: since youre putting the admin in the seeds, you potentially don't need to add it in account (for optimizing size, but this is fine as well)

games/gacha/pinocchio/program/src/state/pull.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Account type discriminator ([`AccountDiscriminator::Pull`]).
+    pub discriminator: u8,
+    /// PDA bump seed.
+    pub bump: u8,
+    /// Current [`PullStatus`].
+    pub status: u8,
+    /// Selected tier index once settled; [`TIER_UNSET`] while pending.
+    pub tier_selected: u8,
+    /// Pool this pull belongs to.
+    pub pool: Address,
```
dev-jodee, line 41: same comment as above, pool, buyer and index oyu might not need to track, since its in your seeds

games/gacha/pinocchio/program/src/ccvrf.rs
```diff
@@ -0,0 +1,186 @@
...
+/// Light registered-program PDA for cc-vrf.
+pub const REGISTERED_PROGRAM_PDA: Address = Address::from_str_const("35hkDgaAKwMCaxRz2ocSZ6NaUrtKkyNqU6c4RV3tYJRh");
+/// Account-compression CPI authority (`["cpi_authority"]` under the Light system program).
+pub const ACCOUNT_COMPRESSION_AUTHORITY: Address =
+    Address::from_str_const("HwXnGK3tPkkVY6P439H2p68AxpeuWXd5PcrAxFpbmfbA");
+/// Light account-compression program.
+pub const ACCOUNT_COMPRESSION_PROGRAM_ID: Address =
+    Address::from_str_const("compr6CUsB5m2jS4Y3831ztGSTnDpnKJTKS95d64XVq");
+/// The canonical Light v2 batched address tree; cc-vrf rejects any other.
+pub const ADDRESS_TREE_V2: Address = Address::from_str_const("amt2kaJA14v3urZbZvnc5v2np8jqvc4Z8zDep5wbtzx");
```
dev-jodee: If there's a crate for cc or light, might be easier to use instead of reimplementing some of their constants + functions)

games/gacha/pinocchio/program/src/instructions/claim_prize.rs
```diff
@@ -0,0 +1,278 @@
...
+/// [`PrizeClaimedEvent`].
+///
+/// Permissionless: any funded signer may crank a claim (the buyer, the operator,
+/// or an indexer). The payer funds the mint and ATA rent; the prize always goes
+/// to the pull's buyer. The mint is a PDA of the pull, so each pull has exactly
+/// one prize, and the mint authority is discarded after minting a supply of one.
+pub fn process(accounts: &mut [AccountView]) -> ProgramResult {
+    let accounts = ClaimPrizeAccounts::try_from(accounts)?;
+
+    let tier;
```
dev-jodee, line 87: code looks good, I might split some of the functionalities into util files, , like all your nft stuff in one util, just to keep the files small and easy to read

dev-jodee (dismissed):

games/gacha/pinocchio/tests/integration-tests/src/utils/constants.rs
amilz: Done. The system, Token-2022, and ATA IDs now come from the crates that declare them (`pinocchio_system::ID`, `pinocchio_token_2022::ID`, `pinocchio_associated_token_account::ID` — all `solana_address::Address` via `declare_id!`, so no conversion), and the program ID comes from the generated client (`gacha_client::GACHA_ID`), which also makes a program/client ID mismatch fail the suite.

The cc-vrf and Light aliases are gone entirely — the tests import `crate::ccvrf::*` directly rather than re-wrapping them, so `constants.rs` is down to four entries.

games/gacha/pinocchio/tests/integration-tests/src/utils/pda.rs
amilz: You were right, and it caught a real gap: `Pull` had **no** `#[codama(seed(...))]` attributes at all, so codama was not generating a derivation for it — that is why it was hand-rolled. Declared them, and the tests now derive through `gacha_client::{Pool,Vault,Pull,PrizeMint}::find_pda`. `utils/pda.rs` is down to `get_ata` (ATA derivation is not in our IDL). Also deleted the hand-written `findPullPda` from the TS client in favour of the generated one.

Following the suggestion also surfaced a codama bug: `@codama/renderers-rust@3.1.1` rendered numeric PDA seeds as `index.to_string()` (decimal ASCII) instead of `to_le_bytes()`, so the generated `Pull::find_pda` disagreed with both the program and the TS client — 17 tests went red. Fixed upstream in 3.1.3; bumped. Worth knowing for any other example with a numeric seed.

games/gacha/pinocchio/SECURITY.md
amilz: Removed. `security_txt!` in `program/src/lib.rs` already points `policy` at the repo-level `SECURITY.md`, so nothing was depending on the copy.

games/gacha/pinocchio/LICENSE
amilz: Removed.

games/gacha/pinocchio/tests/integration-tests/src/utils/constants.rs
amilz: Switched. Worth noting it was a free change: `solana-pubkey` 4.x is a thin alias crate — `pub use solana_address::{Address as Pubkey, ...}` — and everything in the test graph (litesvm 0.12, solana-instruction, solana-message, solana-keypair) already resolves to `solana-address` 2.7.0. So it was a pure rename with no conversions, and the test crate drops `solana-pubkey` entirely and uses the workspace `solana-address`.

games/gacha/pinocchio/tests/integration-tests/src/utils/idl.rs
amilz: It backs `test_account_meta.rs`: for every account the IDL marks writable, the test rebuilds that instruction with the account demoted to read-only and asserts the program rejects it with `AccountNotWritable`. So the writable/signer flags in the IDL are checked against what the program actually enforces, instead of a client silently under-declaring its writes.

Reading it from the committed IDL rather than a hardcoded list means adding an account to an instruction automatically extends the test. Added a doc comment saying this on the function, since it was not obvious from the call site.

games/gacha/pinocchio/program/src/state/common.rs
amilz: Good catch on the failure mode, and agreed it is the right default — but I am leaving the numbering at 0 for this PR, because the fix is not free here.

`Pool`/`Pull` PDAs are `["pool", admin]` and `["pull", pool, buyer, index]`. Renumbering makes every existing account fail `validate()`, and those PDAs cannot be re-initialised: the address is already allocated, so `init_pool` fails at `Allocate`. Recovering means a new admin keypair and a new pool, which strands the live devnet pool (13 pulls) that the deployed webapp points at. For an example whose devnet demo is part of the deliverable, that trade did not seem worth it.

The residual exposure is also narrower than the general case, because `check()` gates on three things, not just the discriminator byte:

```rust
pub fn check(account: &AccountView) -> Result<(), ProgramError> {
    if !account.owned_by(&crate::ID) {
        return Err(GachaError::NotProgramOwned.into());
    }
    Self::validate(&account.try_borrow()?)   // exact-length + discriminator
}
```

`refund_pull` (the only close path) calls `AccountView::close()`, which reassigns the account to the system program — so a closed pull fails the ownership check before the discriminator is ever read, and the exact-length check means a zeroed account of any other size fails too. To exploit it you would need a program-owned, fully-zeroed account of exactly `Pool::LEN`, which nothing in the program produces.

That said, the point stands for anyone copying this as a starting point, so I would rather not leave it implicit. Happy to either (a) start at 1 and reset the devnet pool as a follow-up once the demo is repointed, or (b) land a `README`/`CLAUDE.md` note that new programs should start discriminators at 1. Which would you prefer?

games/gacha/pinocchio/program/src/state/pool.rs
amilz: Keeping `admin`, but the reason is not just size — it is that the pool PDA has to *sign*. `claim_prize` mints the prize with the pool as mint authority, which needs the full seed set:

```rust
let pool_seeds = [Seed::from(POOL_SEED), Seed::from(admin.as_ref()), Seed::from(&pool_bump_bytes[..])];
let pool_signer = [Signer::from(&pool_seeds[..])];
```

`admin` is a seed, so signing as the pool requires it. Dropping the field means passing `admin` as an extra account meta on every instruction where the pool signs or is authorised — `claim_prize`, `withdraw_fees`, `settle_pull` — trading 32 bytes of one-time rent for an extra meta on three instructions, plus a `pool == find_pool_pda(admin)` check in each to make it trustworthy. Net worse on both tx size and validation surface.

Second reason: it keeps the account self-describing. The webapp does `getProgramAccounts` with a `dataSize` filter and reads `admin`/`operator` straight off the decoded pool; without the field it would have to reverse the PDA, which is not possible.

games/gacha/pinocchio/program/src/state/pull.rs
amilz: Same call, and here the fields are load-bearing rather than just convenient:

- `index` is consumed at claim time — it goes into the NFT name (`Gacha Pull #<index>`) and into `PrizeClaimedEvent`. Dropping it means passing it as instruction data and re-deriving the PDA to verify it, which is strictly more work than storing 8 bytes.
- `pool` and `buyer` are what `claim_prize` validates against, so a caller cannot point a settled pull at the wrong pool or redirect the prize:

```rust
if pull.pool != *accounts.pool.address() { return Err(GachaError::PoolMismatch.into()); }
if pull.buyer != *accounts.buyer.address() { return Err(GachaError::BuyerMismatch.into()); }
```

Deriving instead of comparing would work, but it turns two 32-byte compares into a `find_program_address` per check — and `claim_prize` is already the heaviest instruction (mint + metadata + ATA + mint-to + set-authority CPIs).
- `buyer` is also a `memcmp` filter offset the webapp and the operator crank use to list a wallets pulls via `getProgramAccounts`. That only works if it is in the account data.

So the 72 bytes buy a cheaper claim, and indexability. Open to revisiting if you disagree on the tradeoff.

games/gacha/pinocchio/program/src/ccvrf.rs
amilz: Done for the Light half. The five Light addresses now come from `light-sdk-types`:

```rust
use light_sdk_types::constants as light;

pub const LIGHT_SYSTEM_PROGRAM_ID: Address = Address::new_from_array(light::LIGHT_SYSTEM_PROGRAM_ID);
pub const REGISTERED_PROGRAM_PDA: Address = Address::new_from_array(light::REGISTERED_PROGRAM_PDA);
pub const ACCOUNT_COMPRESSION_AUTHORITY: Address = Address::new_from_array(light::ACCOUNT_COMPRESSION_AUTHORITY_PDA);
pub const ACCOUNT_COMPRESSION_PROGRAM_ID: Address = Address::new_from_array(light::ACCOUNT_COMPRESSION_PROGRAM_ID);
pub const ADDRESS_TREE_V2: Address = Address::new_from_array(light::ADDRESS_TREE_V2);
```

All five matched the hand-copied literals exactly, and `.so` size is unchanged since they are compile-time constants.

**cc-vrf has no published crate** (nothing on crates.io under `cc-vrf`/`ccvrf`/`collectorcrypt-*`), so its program ID, CPI authority, and the `commit_proof_with_beta` discriminator + 450-byte wire layout stay hand-written. That part is unavoidable until they publish one.

On which Light crate: I checked all three, and they all bottom out in the same place — `light-sdk` and `light-sdk-pinocchio` both just do `pub use light_sdk_types::{constants, ...}`, so the values are identical whichever you pick. Measured `cargo tree -p gacha-program --edges normal`:

| crate | deps in tree | notes |
| --- | --- | --- |
| `light-sdk-types` | 119 | where the constants are defined |
| `light-sdk-pinocchio` | 128 | pulls **pinocchio 0.9.3** alongside our 0.11.2 (and `pinocchio-system` 0.3.0 + 0.6.1) |
| `light-sdk` | 259 | forces `std` on `light-compressed-account`/`light-hasher`/`light-sdk-types` |

`light-sdk-pinocchio` looks like the natural pick for a pinocchio program, but 0.23.0 is its latest (no 0.24, so it lags the others) and it hard-pins `pinocchio = "0.9"` — two semver-incompatible pinocchio universes with non-interoperable `AccountView`/`Address`. Harmless while we only touch its constants, but a trap for anyone copying this who then reaches for its `cpi::invoke` or `LightAccount`. Since we hand-serialize the cc-vrf CPI anyway and only need five byte arrays, `light-sdk-types` is the honest dependency. Worth switching if Light ships it on pinocchio 0.11+.

games/gacha/pinocchio/program/src/instructions/claim_prize.rs
amilz: Done — `claim_prize.rs` is 279 → 142 lines. All the NFT work moved to `instructions/helpers/prize_nft.rs`: `mint_prize_nft` plus `fund_metadata_rent` / `initialize_metadata` / `set_rarity_field`, and the token-metadata discriminators, `FIELD_KEY_VARIANT`, `RARITY_KEY`, and `MINT_LEN` went with it. `claim_prize` is now validation + status transition + event.

Also took the chance to fix a convention violation in the same directory — `helpers/mod.rs` had the account guards and `create_pda_account` living in it, which our own CLAUDE.md says not to do ("no `mod.rs` business logic"). Split into `helpers/checks.rs` and `helpers/account.rs`, so `mod.rs` is declarations and re-exports only.

dev-jodee (approved):

PR #667, opened 2026-08-05
security: add missing authorization checks in close-account and transfer-sol

NikkiAung:
## Summary

Audited every `native`/`pinocchio` (non-Anchor — no macro-generated account-validation safety net) instruction handler under `basics/` for the classic Solana vulnerability classes: missing signer checks, missing owner checks before deserialization, missing PDA re-derivation, arbitrary CPI, duplicate-account aliasing. Found two real, currently-shipping vulnerabilities. Both are **invisible to the existing test suites** — every existing test only exercises the honest path (the same signer creates and later acts on their own account), so green CI gives false confidence.

This matters more here than in most repos: `program-examples` is code thousands of learners copy-paste directly into real projects, so a vulnerability in a teaching example teaches the vulnerability to everyone who copies it. Both are the textbook "missing signer authorization" pattern — the most common real-world Solana exploit class.

### 1. `basics/close-account` (native + pinocchio): anyone can close and drain anyone else's account

`close_user` / `process_close` took `target_account` and `payer`, and unconditionally moved nearly all of `target_account`'s lamports into `payer`, then reassigned it to the System Program. It never checked `payer.is_signer`, and never re-derived the `["USER", payer.key]` PDA to confirm `target_account` actually belongs to `payer`.

**Exploit:** submit `CloseUser` with `[victim's User PDA, attacker's own pubkey, System Program]`. The attacker doesn't need the victim's key at all — they sign with their own key (satisfying "some signer exists"), pass the victim's PDA as the account to close, and drain it. PDAs are discoverable via `getProgramAccounts`.

The Anchor version of this same example was already safe — `user: Signer<'info>` enforces `is_signer`, and `seeds = [b"USER", user.key().as_ref()]` on `user_account` makes Anchor auto-verify the PDA. Only the native/pinocchio ports dropped both checks when translating away from the framework.

**Fix:** require `payer.is_signer`, and re-derive + compare the expected PDA before touching `target_account`.

### 2. `basics/transfer-sol` (anchor + native + pinocchio — all three): anyone can drain any program-owned account

`transfer_sol_with_program` directly manipulates lamports (`**payer.try_borrow_mut_lamports()? -= amount`) with no CPI and no `is_signer` check anywhere. Since the Solana runtime only requires that the *owning* program initiate a lamport debit (no signature required by the runtime itself), any account owned by this program can be drained by anyone who knows its pubkey.

This is the one case where the bug is **not** an Anchor-vs-native gap — the Anchor version has it too:
```rust
/// CHECK: Use owner constraint to check account is owned by our program
#[account(mut, owner = id())]
payer: UncheckedAccount<'info>,
```
`owner = id()` confirms the account is program-owned — which the runtime requires anyway for the debit to succeed at all, so it doesn't add real protection — but never checks `is_signer`. The top-level `basics/transfer-sol/README.md` presents this as a working pattern with zero caveat about needing an authorization check, which is exactly the shape of bug that has caused real fund-loss incidents in production "vault" programs holding user funds in program-owned accounts.

**Fix:** require `payer.is_signer` in all three implementations. For Anchor, `payer` changes from `UncheckedAccount<'info>` to `Signer<'info>` (constraints like `mut`/`owner` compose with either account type, so this is a minimal, idiomatic change — not adding a redundant owner check to native/pinocchio since the runtime already refuses the debit unless the executing program owns the account).

## Test plan

Every fix ships with a new adversarial test proving the exploit is blocked. Each one was **verified to fail against the unpatched code first** (confirming it actually exercises the bug, not a false negative), then verified to pass after the fix, with every pre-existing legitimate-flow test still passing unchanged:

- [x] `close-account/native`: attacker (own signer, victim's PDA as target) — fails before fix, blocked after. 3/3 tests pass.
- [x] `close-account/pinocchio`: same attack in Rust via `cargo test` — fails before fix, blocked after. 1/1 tests pass.
- [x] `transfer-sol/anchor` (litesvm): attacker marks victim's program-owned account `isSigner: false`, only signs with their own unrelated fee-payer key — fails before fix, blocked after. 3/3 tests pass.
- [x] `transfer-sol/native`: same attack via raw `@solana/kit` account-role manipulation — fails before fix, blocked after. 4/4 tests pass.
- [x] `transfer-sol/pinocchio`: same attack via raw `AccountMeta` — fails before fix, blocked after. 1/1 tests pass. (Caught and fixed a flaw in my first draft of this test: the initial victim account had only ever *received* lamports rather than being created via `create_account`, so it was actually System-Program-owned and the runtime blocked the debit for an unrelated reason. Recreated the victim properly via `create_account` with this program as owner so the test exercises the real vulnerability.)
- [x] `transfer-sol/anchor/tests/test.ts` (the validator-based suite, not litesvm): updated the existing "Transfer SOL with Program" test to add `.signers([payerAccount])`, since `payer` is now `Signer<'info>` and must actually sign — this is a necessary, expected update to the legitimate flow, not a new issue. Could not execute this specific file locally (`solana-test-validator`/`surfpool` hangs at startup in this sandbox — unrelated to the change); the identical on-chain program logic is already fully exercised and verified via the `litesvm.test.ts` suite above, which hits the same compiled program through a full transaction-processing path.

All programs build clean (`cargo check` / `anchor build` / `cargo build-sbf`) with no new warnings beyond pre-existing ones.

### DIFF
--- a/basics/close-account/native/tests/close-account.test.ts
+++ b/basics/close-account/native/tests/close-account.test.ts
@@ -68,6 +68,11 @@ describe('Close Account!', () => {
 
         const result = svm.sendTransaction(signedTx);
         assert(result instanceof FailedTransactionMetadata, 'expected the attacker transaction to fail');
+        assert.include(
+            result.err().toString(),
+            'IncorrectProgramId',
+            `expected the attacker's target PDA to be rejected as not belonging to them, got: ${result.toString()}`,
+        );
     });
 
     it('Close the account', async () => {
--- a/basics/close-account/native/tests/close-account.test.ts
+++ b/basics/close-account/native/tests/close-account.test.ts
@@ -48,6 +48,28 @@ describe('Close Account!', () => {
         assert(!(result instanceof FailedTransactionMetadata), `transaction failed: ${result.toString()}`);
     });
 
+    it("An attacker cannot close another user's account", async () => {
+        // The attacker signs with their own key, but passes the victim's
+        // (payer's) User PDA as the account to close. Without a check that
+        // the target PDA actually belongs to the signer, this would drain
+        // the victim's account into the attacker's.
+        const attacker = await generateKeyPairSigner();
+        svm.airdrop(attacker.address, lamports(1_000_000_000n));
+
+        const ix = createCloseUserInstruction(testAccountAddress, attacker, programId);
+
+        const transactionMessage = pipe(
+            createTransactionMessage({ version: 0 }),
+            m => setTransactionMessageFeePayerSigner(attacker, m),
+            m => svm.setTransactionMessageLifetimeUsingLatestBlockhash(m),
+            m => appendTransactionMessageInstruction(ix, m),
+        );
+        const signedTx = await signTransactionMessageWithSigners(transactionMessage);
+
+        const result = svm.sendTransaction(signedTx);
+        assert(result instanceof FailedTransactionMetadata, 'expected the attacker transaction to fail');
+    });
+
     it('Close the account', async () => {
         const ix = createCloseUserInstruction(testAccountAddress, payer, programId);
 
--- a/basics/close-account/pinocchio/program/tests/tests.rs
+++ b/basics/close-account/pinocchio/program/tests/tests.rs
@@ -53,6 +53,32 @@ fn test_close_account() {
     assert_eq!(account.owner, program_id);
     assert_eq!(&account.data[..5], b"Jacob");
 
+    // An attacker cannot close another user's account: the attacker signs
+    // with their own key, but passes the victim's User PDA as the account
+    // to close. Without a check that the target PDA actually belongs to the
+    // signer, this would drain the victim's account into the attacker's.
+    let attacker = Keypair::new();
+    svm.airdrop(&attacker.pubkey(), LAMPORTS_PER_SOL).unwrap();
+
+    let mut data = Vec::new();
+    data.push(CLOSE_DISCRIMINATOR);
+
+    let attack_ix = Instruction {
+        program_id,
+        accounts: vec![
+            AccountMeta::new(test_account_pubkey, false),
+            AccountMeta::new(attacker.pubkey(), true),
+            AccountMeta::new(solana_system_interface::program::ID, false),
+        ],
+        data,
+    };
+
+    let tx =
+        Transaction::new_signed_with_payer(&[attack_ix], Some(&attacker.pubkey()), &[&attacker], svm.latest_blockhash());
+
+    let res = svm.send_transaction(tx);
+    assert!(res.is_err(), "expected the attacker transaction to fail");
+
     // process_close
     let mut data = Vec::new();
     data.push(CLOSE_DISCRIMINATOR);
--- a/basics/close-account/pinocchio/program/tests/tests.rs
+++ b/basics/close-account/pinocchio/program/tests/tests.rs
@@ -1,9 +1,11 @@
 use litesvm::LiteSVM;
 use solana_instruction::{AccountMeta, Instruction};
+use solana_instruction_error::InstructionError;
 use solana_keypair::{Keypair, Signer};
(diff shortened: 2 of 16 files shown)


basics/transfer-sol/pinocchio/program/tests/test.rs
```diff
@@ -84,4 +84,38 @@ fn test_transfer_sol() {
...
+    let attack_ix = Instruction {
+        program_id,
+        accounts: vec![AccountMeta::new(victim.pubkey(), false), AccountMeta::new(attacker_recipient.pubkey(), false)],
+        data,
+    };
+
+    let tx = Transaction::new_signed_with_payer(&[attack_ix], Some(&payer.pubkey()), &[&payer], svm.latest_blockhash());
+
+    let res = svm.send_transaction(tx);
+    assert!(res.is_err(), "expected the attacker transaction to fail");
```
dev-jodee: should assert on the actual err, not just generalized error

basics/close-account/native/tests/close-account.test.ts
```diff
@@ -48,6 +48,28 @@ describe('Close Account!', () => {
...
+        const transactionMessage = pipe(
+            createTransactionMessage({ version: 0 }),
+            m => setTransactionMessageFeePayerSigner(attacker, m),
+            m => svm.setTransactionMessageLifetimeUsingLatestBlockhash(m),
+            m => appendTransactionMessageInstruction(ix, m),
+        );
+        const signedTx = await signTransactionMessageWithSigners(transactionMessage);
+
+        const result = svm.sendTransaction(signedTx);
+        assert(result instanceof FailedTransactionMetadata, 'expected the attacker transaction to fail');
```
dev-jodee, line 70: should assert on the actual err, not just generalized error

basics/transfer-sol/anchor/tests/litesvm.test.ts
```diff
@@ -93,4 +93,48 @@ describe('LiteSVM: Transfer SOL', () => {
...
+            data,
+        });
+
+        const tx = new Transaction().add(ix);
+        tx.feePayer = payer.publicKey;
+        tx.recentBlockhash = svm.latestBlockhash();
+        tx.sign(payer);
+
+        const result = svm.sendTransaction(tx);
+        assert(result instanceof FailedTransactionMetadata, 'expected the attacker transaction to fail');
```
dev-jodee, line 138: should assert on the actual err, not just generalized error

basics/transfer-sol/native/tests/test.ts
```diff
@@ -95,6 +96,45 @@ describe('transfer-sol', () => {
...
+            accounts: [{ address: victimAccount.address, role: AccountRole.WRITABLE }, ...legitIx.accounts.slice(1)],
+        };
+
+        const transactionMessage = pipe(
+            createTransactionMessage({ version: 0 }),
+            m => setTransactionMessageFeePayerSigner(payer, m),
+            m => svm.setTransactionMessageLifetimeUsingLatestBlockhash(m),
+            m => appendTransactionMessageInstructions([attackIx], m),
+        );
+        const signedTx = await signTransactionMessageWithSigners(transactionMessage);
```
dev-jodee, line 133: should assert on the actual err, not just generalized error

basics/close-account/pinocchio/program/tests/tests.rs
```diff
@@ -53,6 +53,36 @@ fn test_close_account() {
...
+        data,
+    };
+
+    let tx = Transaction::new_signed_with_payer(
+        &[attack_ix],
+        Some(&attacker.pubkey()),
+        &[&attacker],
+        svm.latest_blockhash(),
+    );
+
```
dev-jodee, line 84: should assert on the actual err, not just generalized error

dev-jodee (changes requested):
thx ! all the same comments, assert on real error instead of just "an error happened"

dev-jodee (approved):

PR #668, opened 2026-08-05
security: fix broken take_offer invariant + add missing RefundOffer instruction

NikkiAung:
## Summary

Escalated the same security-audit methodology from #667 (missing signer checks in `basics/`) to `tokens/escrow` — native, pinocchio, and anchor. Escrow/vault programs are the highest-value target for this kind of audit: they hold other people's funds in program custody, and this repo's version is what learners copy into real token-swap escrows.

### Bug (native only): broken post-transfer invariant lets a third party permanently freeze any offer

`take_offer`'s post-transfer sanity check compared the maker's token-B balance against the wrong variable:

```rust
assert_eq!(maker_amount_b, taker_amount_a_before_transfer + offer.token_b_wanted_amount); // BUG
```

`taker_amount_a_before_transfer` is a copy-paste of the line above, only half-edited — it should be `maker_amount_b_before_transfer`. The real SPL Token CPI always moves the correct amount, so this assert only happened to pass when the maker's token-B balance was exactly 0 before the trade. That's not guaranteed, and is trivially breakable by **any third party, no cooperation needed**: create the maker's token-B ATA (permissionless) and send it 1 base unit (also permissionless — no signature required from the recipient) before a `take_offer` call. From that point on, every `take_offer` for that offer panics and reverts.

### Gap (all three implementations): no way for a maker to reclaim their deposit

This is what turns the bug above from "annoying" into **permanent fund loss**: there was no `RefundOffer`/cancel instruction anywhere — not in native, not in pinocchio, not even in the anchor reference version. Only `MakeOffer`/`TakeOffer` existed. Once an offer is bricked (by the bug above, or simply because the counterparty never shows up — the ordinary non-adversarial case), the maker's deposited tokens were locked in the vault with no recovery path.

## Fixes

- One-line fix to the comparison in `native/program/src/instructions/take_offer.rs`.
- New `RefundOffer` instruction in **all three** implementations: only the maker (verified against the offer's stored `maker` field) can call it. Returns the vault's current token-A balance to the maker's own account, closes the vault and offer accounts, rent to the maker. Structurally mirrors each implementation's existing `take_offer` vault-drain-and-close logic — same seeds, same CPI shape — just redirecting the destination and dropping the token-B leg entirely (a refund never touches token B).

## Test plan

Every new test was verified to **fail against the unpatched/pre-feature code first** (confirming it actually exercises the bug/gap, not a false negative), then pass after the fix — same discipline as #667:

- [x] **native**: regression test pre-funds the maker's token-B account before `Take Offer` to reproduce the exact condition that broke the old assert (confirmed panic at `take_offer.rs:171` before the fix); `RefundOffer` happy-path and non-maker-cannot-refund tests. 6/6 passing.
- [x] **pinocchio**: same two `RefundOffer` tests (never had the assert bug — its `take_offer` has no equivalent check). 4/4 passing.
- [x] **anchor**: same two `RefundOffer` tests added to `litesvm.test.ts`. 4/4 passing. (The separate validator-based `escrow.test.ts` — pre-existing, untouched — couldn't be executed in this sandbox due to a local `solana-test-validator` startup issue unrelated to this change; the identical on-chain logic is already fully exercised via `litesvm.test.ts` against the same compiled program.)
- [x] `cargo fmt --check` and `cargo clippy -- -D warnings` clean for the root-workspace crates (native + pinocchio); `prettier --check` clean for all touched TS files. (`tokens/escrow/anchor` is its own nested Cargo workspace, outside the root workspace `cargo fmt`/`clippy` cover — consistent with the rest of the repo's `*/anchor/` programs.)
- [x] Added `#[allow(clippy::enum_variant_names)]` to native's `EscrowInstruction` enum — adding the third `RefundOffer` variant tripped the lint (it only fires at ≥3 variants), and the shared `Offer` postfix is intentional domain vocabulary matching the existing `MakeOffer`/`TakeOffer` naming, not something to rename.

## Incidental fix

Extended native's and pinocchio's test `createValues()` helper to actually respect all the overridable defaults its own type signature (`TestValuesDefaults`) already promised — it previously only read `programId`/`id` from the passed defaults and silently regenerated `maker`/`taker`/mint keypairs regardless, which made it impossible to create a second offer reusing the same already-funded accounts and already-deployed program (needed for the new tests, which create additional offers under the same setup).

### DIFF
--- a/tokens/escrow/native/program/src/instructions/take_offer.rs
+++ b/tokens/escrow/native/program/src/instructions/take_offer.rs
@@ -70,6 +70,11 @@ impl TakeOffer {
         assert_is_associated_token_account(maker_token_account_b.key, maker.key, token_mint_b.key)?;
         assert_is_associated_token_account(taker_token_account_a.key, taker.key, token_mint_a.key)?;
 
+        // validate the vault is the offer's actual vault, not a substitute
+        // token-A account that also happens to be owned by the offer PDA
+        //
+        assert_is_associated_token_account(vault.key, offer_info.key, token_mint_a.key)?;
+
         // create taker token A account if needed, before receiveing tokens
         //
         if taker_token_account_a.lamports() == 0 {
--- a/tokens/escrow/native/program/src/instructions/take_offer.rs
+++ b/tokens/escrow/native/program/src/instructions/take_offer.rs
@@ -168,7 +168,7 @@ impl TakeOffer {
         let maker_amount_b = TokenAccount::unpack(&maker_token_account_b.data.borrow())?.amount;
 
         assert_eq!(taker_amount_a, taker_amount_a_before_transfer + vault_amount_a);
-        assert_eq!(maker_amount_b, taker_amount_a_before_transfer + offer.token_b_wanted_amount);
+        assert_eq!(maker_amount_b, maker_amount_b_before_transfer + offer.token_b_wanted_amount);
 
         let taker_amount_b = TokenAccount::unpack(&taker_token_account_b.data.borrow())?.amount;
         let vault_amount_a = TokenAccount::unpack(&vault.data.borrow())?.amount;
--- a/tokens/escrow/pinocchio/program/src/instructions/refund_offer.rs
+++ b/tokens/escrow/pinocchio/program/src/instructions/refund_offer.rs
@@ -52,6 +52,16 @@ pub fn refund_offer(program_id: &Address, accounts: &mut [AccountView], _data: &
         return Err(ProgramError::InvalidSeeds);
     }
 
+    // Verify vault is the offer's actual vault, not a substitute token-A
+    // account that also happens to be owned by the offer PDA.
+    let (expected_vault, _) = Address::find_program_address(
+        &[offer_account.address().as_ref(), pinocchio_token::ID.as_ref(), token_mint_a.address().as_ref()],
+        &pinocchio_associated_token_account::ID,
+    );
+    if vault.address() != &expected_vault {
+        return Err(ProgramError::InvalidAccountData);
+    }
+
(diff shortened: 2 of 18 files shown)


tokens/escrow/native/program/src/instructions/take_offer.rs
dev-jodee, line 170: before using vault auth, can we add assert_is_associated_token_account(vault, offer_info,  token_mint_a)

tokens/escrow/pinocchio/program/src/instructions/refund_offer.rs
```diff
@@ -0,0 +1,100 @@
...
+/// Refunds an open offer: the vaulted token A is returned to the maker and
+/// the vault and offer accounts are closed. Only the maker who created the
+/// offer may call this.
+///
+/// Accounts:
+///   0. `[writable]`         offer account (PDA `[b"offer", maker, id]`, closed here)
+///   1. `[]`                 token mint A
+///   2. `[writable]`         maker's token A account (receives the refund)
+///   3. `[writable]`         vault (offer PDA's token A account, drained and closed)
+///   4. `[signer, writable]` maker
```
dev-jodee, line 19: can we do same check in take_offer.rs (find_program_address)

tokens/escrow/pinocchio/program/src/instructions/refund_offer.rs
```diff
@@ -0,0 +1,100 @@
+use pinocchio::{
```
dev-jodee, line 1: add canonical check maker_token_account_a

dev-jodee (changes requested):

dev-jodee (approved):

PR #672, opened 2026-08-05
fix(abl-token): block wallets from sending, not just receiving

NikkiAung:
## Summary

`allow-block-list-token`'s transfer hook only ever checked the **destination** wallet's allow/block status. A wallet explicitly blocked via `init_wallet` (`ABWallet { allowed: false }`) could still send its entire balance out through an ordinary `TransferChecked` to any unlisted destination — in every mint mode (Allow/Block/Threshold), since the sender's `ABWallet` record was never read at all.

This directly contradicts the `WalletBlocked` error name and the "Blocked" badge shown in the admin UI, and defeats the standard real-world reason to build a block-list token (e.g. sanctions compliance: stopping the sanctioned party from moving funds *out*).

**Root cause:** `utils.rs`'s `get_extra_account_metas()` configured exactly one extra account for the hook's `Execute` call, seeded from `account_index: 2` (the destination token account). `tx_hook.rs` had a `source_token_account` field on the accounts struct but never read it — the decision logic never had access to the sender's status in any mode.

## Fix

- `utils.rs` — add a second extra account seeded from the source token account's owner (`account_index: 0`), so Token-2022 resolves and appends both the source's and destination's `ab_wallet` PDAs to every `Execute` CPI.
- `tx_hook.rs` — add `source_ab_wallet` to the `TxHook` accounts struct (ordered to match `get_extra_account_metas()`), and extend the decision logic so a wallet with an explicit `allowed: false` record is rejected regardless of which side of the transfer it's on. Allow/Threshold mode's documented "who may *receive*" semantics (see this example's README) are left untouched — only the block-list's own "blocked" concept, whose name and UI already imply full blocking, becomes bidirectional.
- The decision matrix is pulled out into a standalone `decide(mint_mode, source_mode, destination_mode, amount)` function with unit tests, since this program has no integration-test harness that exercises `tx_hook` via a real hooked transfer (`tests/test.rs` only covers `init_config`/`init_mint`).
- `src/components/account/account-data-access.tsx` — the demo frontend's `useSendTokens()` manually constructs the hook's extra accounts rather than relying on Token-2022's standard client-side auto-resolution. Updated it to push both the source and destination `ab_wallet` PDAs, in the same order the program now expects, so the demo UI stays consistent with the hardened program. This is a demo-consistency fix, not itself security-critical — any standards-conformant Token-2022 client picks up the corrected on-chain `ExtraAccountMetaList` automatically. Not verified via live browser/wallet testing (no validator available in this environment); it's a direct structural mirror of the pre-existing destination-account push, keyed on the sender's own public key instead.

## Out of scope

`block-list/pinocchio` (a separate, unrelated example) has a self-documented limitation in its own README where setting up the `ExtraAccountMetaList` while the block list is still empty locks in "0 extra accounts" until `setup-extra-metas` is manually re-run after the first block. Lower severity (requires operator error, not silently wrong-by-design) and a different program — worth a follow-up PR, not bundled here.

## Test plan

- [x] `cargo check` — `abl-token` compiles clean with the new account and function signature.
- [x] `cargo test --lib` — 9/9 pass, including the new `decide()` unit tests covering both sides of the block check and the existing Allow/Block/Threshold semantics.
- [x] Verified the key test actually catches the original bug: temporarily reintroduced the old (destination-only) logic and confirmed `source_blocked_is_always_rejected` fails against it, then confirmed all 9 tests pass again against the fix.
- [x] `anchor build` — IDL regenerates with exactly 7 accounts for `tx_hook`, in the order Token-2022 will actually append them.
- [x] `pnpm exec tsc --noEmit` on the frontend — clean, no type errors from the `account-data-access.tsx` change.
- [x] `pnpm run lint` / `pnpm run format:check` — clean (pre-existing warnings/diffs in untouched files left as-is).
- [ ] Live end-to-end test of the frontend send flow against a real validator (not available in this environment).

### DIFF
--- a/tokens/token-2022/transfer-hook/allow-block-list-token/anchor/programs/abl-token/src/instructions/tx_hook.rs
+++ b/tokens/token-2022/transfer-hook/allow-block-list-token/anchor/programs/abl-token/src/instructions/tx_hook.rs
@@ -24,7 +24,9 @@ pub struct TxHook<'info> {
     /// CHECK:
     pub meta_list: UncheckedAccount<'info>,
     /// CHECK:
-    pub ab_wallet: UncheckedAccount<'info>,
+    pub source_ab_wallet: UncheckedAccount<'info>,
+    /// CHECK:
+    pub destination_ab_wallet: UncheckedAccount<'info>,
 }
 
 impl TxHook<'_> {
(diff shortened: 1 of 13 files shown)


tokens/token-2022/transfer-hook/allow-block-list-token/anchor/programs/abl-token/src/instructions/tx_hook.rs
```diff
@@ -106,14 +95,132 @@ impl TxHook<'_> {
...
     Block,
     None,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn source_blocked_is_always_rejected() {
```
dev-jodee, line 152: dont think the ci is running those test, might need 

cargo test -p abl-token

tokens/token-2022/transfer-hook/allow-block-list-token/anchor/programs/abl-token/src/utils.rs
```diff
@@ -7,12 +7,27 @@ use spl_tlv_account_resolution::{
 use crate::AB_WALLET_SEED;
 
 pub fn get_meta_list_size() -> Result<usize> {
-    Ok(ExtraAccountMetaList::size_of(1).map_err(|_| ProgramError::InvalidArgument)?)
+    Ok(ExtraAccountMetaList::size_of(2).map_err(|_| ProgramError::InvalidArgument)?)
```
dev-jodee: can we have a const for that magic number

tokens/token-2022/transfer-hook/allow-block-list-token/anchor/programs/abl-token/tests/test.rs
```diff
@@ -98,9 +105,151 @@ fn test() {
...
-    let tx = Transaction::new(&[&admin_kp, &mint_kp], msg, svm.latest_blockhash());
+    let tx = Transaction::new(&[admin_kp, &mint_kp], msg, svm.latest_blockhash());
+
+    svm.send_transaction(tx).unwrap();
+
+    (mint_pk, meta_list)
+}
+
+#[test]
+fn test() {
```
dev-jodee: can we have a better test name

dev-jodee (changes requested):
don't think you ran cargo fmt as well

dev-jodee (approved):

PR #674, opened 2026-08-06
security: fix inverted fundraiser deadline logic in token-fundraiser

NikkiAung:
## Summary

`token-fundraiser`'s time-based access control is inverted, breaking the program's core promise (contribute while active, refund if the goal isn't met by the deadline) and permanently locking contributor funds in the realistic case.

**`contribute.rs`** required `duration <= elapsed_days` to pass — contributions only succeeded *after* the fundraiser's duration had already elapsed, staying open forever after that. While the fundraiser is genuinely "active," every contribution reverts with `FundraiserEnded` — the opposite of what that error name implies.

**`refund.rs`** required `duration >= elapsed_days` to pass — refunds only succeeded *before* the deadline. Once the deadline genuinely passes (exactly when `contribute()` *starts* working per the bug above), `refund()` stops working.

**Net effect:** with any realistic nonzero duration, `contribute()` and `refund()` are live in mutually-exclusive, backwards windows. Nobody can contribute while the fundraiser claims to be running; once contributions start flowing in (only possible post-deadline), refund is permanently blocked. If the target isn't met, funds sit in the vault with **no instruction able to move them** — `check_contributions` requires the target met, `refund` requires "not yet past deadline," which is now false by construction. A fund-lock, not a theft, but a complete break of the contract's core guarantee.

**Confirmed against intent, not just guessed:** the example's own `readme.MD` prose says contribute "checks that the fundraising duration has not elapsed" and refund is for "if the duration... has elapsed" — describing the correct behavior while the code implements the opposite. More directly: the README's refund code snippet *already has the correct form* — it's the actual `refund.rs` source that drifted from the README, not the other way around. The README's contribute snippet has the same inverted form as the real bug and needed the identical fix.

**Why no test caught this:** both test suites called `initialize(..., 0)` — duration=0. With duration=0, `elapsed_days >= 0` is always true, which trivially (and accidentally) satisfies *both* inverted checks at once, completely masking the inversion. The "robustness test" cases also swallowed errors without asserting anything.

## Fix

Minimal, two-character operator flips (not expression rewrites), so the diff mirrors the README's already-correct refund form:
- `contribute.rs`: `duration <= elapsed` → `duration > elapsed` (window still open).
- `refund.rs`: `duration >= elapsed` → `duration <= elapsed` (window has closed).
- `readme.MD`: same flip for the one inverted snippet (contribute section); the refund section was already correct, left untouched.
- Fixed the misleading comments above both checks ("Check if the fundraising duration has been reached" described neither direction correctly).

Verified boundary semantics precisely: at `elapsed_days == duration`, contribute must be closed and refund must be open — a deadline is an exclusive upper bound for contributing.

## Test changes

Both suites called `initialize(..., 0)`. Post-fix, `duration = 0` means "expired at creation" (was "never expires" pre-fix) — no validation exists on `duration` in `initialize`, so this is a real semantic change worth flagging, not just a test-parameter tweak. Both suites needed a realistic nonzero duration to exercise anything meaningful.

- **`tests/litesvm.test.ts`** (can deterministically warp its own clock): added a test that warps to the *exact* deadline boundary and confirms `contribute()` is rejected right at that instant (a direct, isolated repro of the "only works after the deadline" bug — this is the cleanest single proof that the fix is correct). Added a test confirming `refund()` is rejected while still active. Moved the existing happy-path refund test to run after the boundary warp, with stronger assertions (contributor ATA balance restored, vault drained to zero, `Contributor` account closed). Tightened the "robustness" tests to assert specific error codes instead of swallowing errors.
- **`tests/fundraiser.ts`** (real `solana-test-validator`, no way to fast-forward its clock — confirmed there's no RPC for it and `--warp-slot` is a startup-only flag that doesn't help mid-suite): restructured the final refund test into an assertion that refund is correctly rejected while still active, with a vault-balance-unchanged check. Tightened the two "robustness" tests the same way.

**Honesty note on test design:** with a realistic nonzero duration, `contribute()` was broken from its very first call pre-fix (confirmed by running the new tests against the unpatched code), so the "refund rejected while active" tests actually fail pre-fix via a *cascading* `AccountNotInitialized` (no `Contributor` account ever got created) rather than a direct "refund wrongly succeeds" repro — that specific symptom only reproduced under the original tests' degenerate `duration=0` setup. Both are still valid regression tests (fail before the fix, pass after with the specific intended error), and the litesvm boundary test independently and cleanly proves the `contribute()` half in isolation.

## Verification

Ran the full pre-fix/post-fix × old-tests/new-tests matrix locally (both suites, matching what `anchor test --validator legacy` runs in CI):
- Pre-fix + original tests (duration=0): all green — reproduces exactly how the masking happened.
- Pre-fix + new tests (duration=1): multiple independent failures, including the boundary test directly showing contribute() succeeding when it should have rejected.
- Post-fix + new tests: all 16 tests pass across both suites.

Also ran `cargo check`, `pnpm exec tsc --noEmit`, and `prettier --check .` (root, which is what CI's Prettier job actually runs — the subproject's own `pnpm lint` script pins an unrelated, older prettier version and isn't used by CI). No IDL/client regeneration needed — only `require!` conditions changed, not account structs or instruction signatures.

## Explicitly out of scope (flagged as follow-ups, not fixed here)

- **Lossy cast on a possibly-negative elapsed value:** `as u16` on a negative `i64` (clock set backwards) wraps to a large u16, which would flip both checks' effective behavior. A hardened version would widen to `i64` with `saturating_sub` instead of narrowing. Left out to keep this a minimal, obviously-correct diff.
- **`refund.rs`/`checker.rs` gate on `vault.amount` (the live ATA balance) instead of `fundraiser.current_amount` (the tracked value).** Since the vault is a plain ATA, anyone can transfer tokens into it directly to push `vault.amount >= amount_to_raise`, blocking refunds or enabling `check_contributions` without genuine contributions. Real, separate finding.
- **`check_contributions` closes the `Fundraiser` PDA while `Contributor` PDAs may still exist**, after which `refund` can never run for those contributors, stranding their rent. Also separate.

### DIFF
--- a/tokens/token-fundraiser/anchor/tests/litesvm.test.ts
+++ b/tokens/token-fundraiser/anchor/tests/litesvm.test.ts
@@ -12,11 +12,27 @@ import {
 import { PublicKey } from '@solana/web3.js';
 import { LiteSVMProvider } from 'anchor-litesvm';
 import BN from 'bn.js';
+import { assert } from 'chai';
 import { LiteSVM } from 'litesvm';
 import IDL from '../target/idl/fundraiser.json';
 import type { Fundraiser } from '../target/types/fundraiser';
 
 const PROGRAM_ID = new PublicKey(IDL.address);
+const SECONDS_PER_DAY = 86400n;
+
+// Asserts that `promise` rejects with the given Anchor custom error code
+// (e.g. 'FundraiserNotEnded'), not just "something failed" - see the same
+// helper in tests/fundraiser.ts for why this matters.
+const expectAnchorError = async (promise: Promise<unknown>, code: string) => {
+    let caught: any;
+    try {
+        await promise;
+    } catch (error) {
+        caught = error;
+    }
+    assert.isDefined(caught, `expected the transaction to fail with ${code}`);
+    assert.strictEqual(caught?.error?.errorCode?.code, code, `expected ${code}, got: ${caught}`);
+};
 
 describe('fundraiser litesvm', () => {
     const client = new LiteSVM();
@@ -78,8 +94,13 @@ describe('fundraiser litesvm', () => {
     it('Initialize Fundaraiser', async () => {
         const vault = getAssociatedTokenAddressSync(mint, fundraiser, true);
 
+        // duration=1 (day). Unlike fundraiser.ts, this suite can warp its
+        // own clock deterministically, so it covers the full lifecycle:
+        // contribute while active, refund rejected while active, contribute
+        // rejected once the deadline passes, and refund succeeding once it
+        // has (see the tests below, in that order).
         const tx = await program.methods
-            .initialize(new BN(30000000), 0)
+            .initialize(new BN(30000000), 1)
             .accountsPartial({
                 maker: maker.publicKey,
                 fundraiser,
@@ -143,10 +164,16 @@ describe('fundraiser litesvm', () => {
     });
 
     it('Contribute to Fundraiser - Robustness Test', async () => {
-        try {
-            const vault = getAssociatedTokenAddressSync(mint, fundraiser, true);
+        // Contributor already holds 2_000_000, and the per-contributor cap
+        // is 10% of the 30_000_000 target = 3_000_000. This 2_000_000
+        // attempt would push the total to 4_000_000, over the cap. Must run
+        // before the deadline warp below - once the deadline passes,
+        // contribute() rejects on the time check first, which would test
+        // the wrong thing.
+        const vault = getAssociatedTokenAddressSync(mint, fundraiser, true);
 
-            const tx = await program.methods
+        await expectAnchorError(
+            program.methods
                 .contribute(new BN(2000000))
                 .accountsPartial({
                     contributor: provider.publicKey,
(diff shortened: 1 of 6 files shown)


tokens/token-fundraiser/anchor/tests/litesvm.test.ts
```diff
@@ -12,11 +12,27 @@ import {
...
 import IDL from '../target/idl/fundraiser.json';
 import type { Fundraiser } from '../target/types/fundraiser';
 
 const PROGRAM_ID = new PublicKey(IDL.address);
+const SECONDS_PER_DAY = 86400n;
+
+// Asserts that `promise` rejects with the given Anchor custom error code
+// (e.g. 'FundraiserNotEnded'), not just "something failed" - see the same
+// helper in tests/fundraiser.ts for why this matters.
+const expectAnchorError = async (promise: Promise<unknown>, code: string) => {
```
dev-jodee: should be exported in a util function instead of reimplemented

tokens/token-fundraiser/anchor/tests/litesvm.test.ts
```diff
@@ -156,22 +183,55 @@ describe('fundraiser litesvm', () => {
...
-            console.log(error.msg);
-        }
+                .rpc(),
+            'MaximumContributionsReached',
+        );
+    });
+
+    // Confirms refund() is correctly gated on FundraiserNotEnded while the
+    // fundraiser is genuinely still active. Note this doesn't reproduce the
+    // original bug in isolation: with a realistic nonzero duration,
```
dev-jodee: please try to reduce the size of the comments, LLM are pretty verbose but this just makes the code really hard to read

dev-jodee (changes requested):
two comments, across the file please reduce comment size it makes the code bloated

dev-jodee (approved):

PR #685, opened 2026-08-10
fix(tokens/token-fundraiser): bind check_contributions to the recorded mint

moviendome:
## Summary

In `tokens/token-fundraiser` (anchor), the `check_contributions` instruction
does not bind the mint to the one recorded when the campaign was created.

`contribute` and `refund` both constrain the fundraiser account with
`has_one = mint_to_raise`, so the mint account passed in the transaction has to
equal the `mint_to_raise` stored at `initialize`. `checker.rs` is the one
instruction of the three without that constraint:

```rust
#[account(
    mut,
    seeds = [b"fundraiser".as_ref(), maker.key().as_ref()],
    bump = fundraiser.bump,
    close = maker,
)]
pub fundraiser: Account<'info, Fundraiser>,
```

Its `mint_to_raise` — and the `vault` derived from it via
`associated_token::mint = mint_to_raise` — are therefore whatever the caller
supplies, kept self-consistent with each other but not tied to the campaign.
The `vault.amount >= amount_to_raise` goal check then measures an unrelated
token account, and because the same instruction carries `close = maker`, a call
made with a substitute mint also closes the `Fundraiser` account — the state
`refund` needs in order to return real contributors' deposits.

## Fix

Add the same `has_one = mint_to_raise` the two sibling instructions already
carry:

```rust
#[account(
    mut,
    seeds = [b"fundraiser".as_ref(), maker.key().as_ref()],
    bump = fundraiser.bump,
    has_one = mint_to_raise,
    close = maker,
)]
pub fundraiser: Account<'info, Fundraiser>,
```

## Test

Adds `tests/checker-mint-binding.test.ts`, a self-contained litesvm spec that:

1. initializes a campaign against one mint,
2. funds the fundraiser's ATA for a *different* mint up to the goal, and
3. calls `check_contributions` with that different mint.

It asserts the call is rejected and the campaign account still exists. Without
the constraint the call succeeds and the campaign is closed; with it, Anchor
rejects the mismatch (`ConstraintHasOne`) before the handler runs.

Verified red→green against the fix.

### DIFF
--- a/tokens/token-fundraiser/anchor/tests/checker-mint-binding.test.ts
+++ b/tokens/token-fundraiser/anchor/tests/checker-mint-binding.test.ts
@@ -10,25 +10,16 @@ import {
 } from '@solana/spl-token';
 import { PublicKey } from '@solana/web3.js';
 import { LiteSVMProvider } from 'anchor-litesvm';
-import { assert } from 'chai';
+import { assert, expect, use } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
 import { LiteSVM } from 'litesvm';
 import IDL from '../target/idl/fundraiser.json';
 import type { Fundraiser } from '../target/types/fundraiser';
 
+use(chaiAsPromised);
+
 const PROGRAM_ID = new PublicKey(IDL.address);
 
-// Regression test for the missing mint binding on `checker.rs`.
-//
-// `contribute` and `refund` both carry `has_one = mint_to_raise` on the
-// `fundraiser` account, so the mint supplied in the transaction must equal the
-// one recorded at `initialize`. `check_contributions` omits that constraint, so
-// its `mint_to_raise` (and the vault derived from it) is whatever the caller
-// passes. A maker can therefore satisfy the goal check against a throwaway mint
-// they control and trigger `close = maker`, destroying the real campaign that
-// every contributor's refund depends on.
-//
-// With the constraint present, Anchor rejects the wrong mint before the handler
-// runs and the campaign account survives.
 describe('fundraiser checker mint binding', () => {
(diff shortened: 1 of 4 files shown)


tokens/token-fundraiser/anchor/tests/checker-mint-binding.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+        const fundTx = new anchor.web3.Transaction().add(
+            createAssociatedTokenAccountInstruction(maker.publicKey, fakeVault, fundraiser, fakeMint),
+            createAssociatedTokenAccountInstruction(maker.publicKey, makerFakeAta, maker.publicKey, fakeMint),
+            createMintToInstruction(fakeMint, fakeVault, maker.publicKey, AMOUNT_TO_RAISE),
+        );
+        await provider.sendAndConfirm(fundTx, [maker]);
+
+        // Call the payout instruction with the fake mint and its funded vault.
+        let rejected = false;
```
dev-jodee: try catch for "finding an error", isn't the proper way to do it, you should actually assert that the error is present, else your assertion is too vague

dev-jodee (changes requested):

tokens/token-fundraiser/anchor/tests/checker-mint-binding.test.ts
moviendome: Thanks for the feedback!

Added `chai-as-promised` and the test now asserts the call rejects with the specific `AnchorError (ConstraintHasOne)` instead of catching anything.

Followed the pattern already used in `tokens/token-2022/transfer-hook/counter`

dev-jodee (approved):

PR #687, opened 2026-08-13
refactor(basics/close-account): replace borsh with kit codecs in native example

amilz:
## Summary
- `basics/close-account/native` was the last package pinned to `borsh: ^0.7.0` — the only pre-2.x pin in the repo — and called `deserialize(schema, Class, buffer)`, a signature that no longer exists in borsh 2.x. Dropped the dep and regenerated the lockfile.
- Instruction data now uses `getStructEncoder`, with `addEncoderSizePrefix(getUtf8Encoder(), getU32Encoder())` for the `name` string, matching the pattern from #676.
- `ts/state/index.ts` exposes a `userDecoder` for the on-chain `User` layout in place of the borsh class boilerplate.
- Test asserts real post-state: the decoded account `name` after create, and the System Program owner plus zero-length data after close.

## Test Plan
- `pnpm build-and-test` in `basics/close-account/native` — 3 passing
- `pnpm exec tsc --noEmit -p tsconfig.json` — clean
- `pnpm run check` (prettier) — clean
- Broke each new assertion once to confirm it fails loudly

Closes DEV-838

### DIFF
--- a/basics/close-account/native/package.json
+++ b/basics/close-account/native/package.json
@@ -8,8 +8,7 @@
   },
   "dependencies": {
     "@solana/kit": "^7.0.0",
-    "@solana-program/system": "^0.13.0",
-    "borsh": "^0.7.0"
+    "@solana-program/system": "^0.13.0"
   },
   "devDependencies": {
     "@types/chai": "^5.2.3",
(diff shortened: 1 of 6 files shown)


dev-jodee (approved):
Refactor borsh→kit codecs in close-account. All CI green.

PR #690, opened 2026-08-21
security: fix swapped transfer amounts + broken invariant check in token-swap

NikkiAung:
## Summary

`swap_exact_tokens_for_tokens`'s `swap_a == false` branch (B→A swaps) has two independent bugs in the same ~20-line block. Neither was ever exercised by the existing tests — the only swap test always calls `swap_a = true`.

### Bug 1 (the severe one): transfer amounts are swapped in the B→A branch

```rust
} else {
    token::transfer(/* pool_account_a -> trader_account_a */ .., input)?;   // should be `output`
    token::transfer(/* trader_account_b -> pool_account_b */ .., output)?;  // should be `input`
}
```
Compare to the correct `swap_a == true` branch just above it, which moves `input` trader→pool and `output` pool→trader. In the `else` branch the two amounts are swapped: the pool pays out `input` (the raw, caller-chosen B amount, reused as if it were the A payout) instead of the formula-computed `output`, and the trader only pays `output` instead of the full `input` they claimed to be swapping.

**Hand-verified this drains real value**: on a pool with `pool_a=1,000,000` / `pool_b=4,000,000` (5% fee), `input=500,000` lets a trader pull out half the pool's entire A reserve while only paying ~106,145 B — the invariant drops from `4e12` to `~2.05e12` in one call.

### Bug 2: the post-trade invariant check compares the wrong accounts

```rust
if invariant > ctx.accounts.pool_account_a.amount * ctx.accounts.pool_account_a.amount {
    return err!(TutorialError::InvariantViolated);
}
```
`pool_account_a.amount` is multiplied by **itself** instead of by `pool_account_b.amount` — comparing pre-trade `A0*B0` against post-trade `A1*A1`, not `A1*B1`. This check is meant to be the AMM's final safety net (per its own comment) but as written never actually verifies the real constant-product invariant.

### Why both had to be fixed together

These two bugs currently mask each other by coincidence: severe drains via Bug 1 always shrink `pool_account_a` a lot, and shrinking `A` a lot also shrinks `A²` a lot — so Bug 2's check, wrong as it is, still happens to reject the most extreme drain attempts. A more moderate trade can still be profitably exploitable while slipping under that coincidental threshold. Fixing either bug alone — without the other — would remove the accidental protection and make the remaining bug fully exploitable. Both are fixed in this PR.

### Secondary fix: overflow-safe arithmetic

Also switched the pre/post-trade invariant computation and the fee calculation to `u128` arithmetic, matching the checked-math style already used for the output-amount calculation in the same function. The raw `u64` multiplications could otherwise overflow and panic (`overflow-checks = true` in this workspace's release profile) for large-supply pools.

## Test changes

Added `'Swap from B to A'` to `tests/swap.ts` — the first test in this example to exercise that branch at all. Verified it fails against the unpatched code (reverts with `InvariantViolated`, since the trade genuinely destroys value under the old code) and passes after the fix, with all other existing tests unaffected.

## Verification

- `cargo check` clean.
- `anchor build` — no account-struct or instruction-signature changes, IDL unaffected beyond recompilation.
- Full `anchor test` suite (`create-amm.ts`, `create-pool.ts`, `deposit-liquidity.ts`, `withdraw-liquidity.ts`, `swap.ts`) — all 10 tests pass post-fix.
- Ran the pre-fix/post-fix matrix: captured the fix as a patch, reverted only the program file, rebuilt, confirmed the new test fails (and only that test — nothing else regresses), reapplied the patch, confirmed all 10 pass.
- `pnpm exec tsc --noEmit` and `prettier --check` clean on the modified test file.

## Explicitly out of scope

- `create_amm.rs`'s `admin` field is never actually used to gate any instruction (dead authority field) — a minor design smell, not exploitable since nothing checks it, not touched here.
- Pool creation is fully permissionless (anyone can create a pool for any AMM+mint pair) — standard, intentional AMM-factory design, not a bug.
- `deposit_liquidity.rs`'s `MINIMUM_LIQUIDITY` handling (withheld from the first LP mint rather than minted-then-burned) — investigated and confirmed self-consistent with `withdraw_liquidity.rs`'s redemption formula, which compensates for it in the supply denominator. Not a bug.

### DIFF
--- a/tokens/token-swap/anchor/programs/token-swap/src/instructions/swap_exact_tokens_for_tokens.rs
+++ b/tokens/token-swap/anchor/programs/token-swap/src/instructions/swap_exact_tokens_for_tokens.rs
@@ -26,10 +26,10 @@ pub fn swap_exact_tokens_for_tokens(
     };
 
     // Apply trading fee, used to compute the output.
-    // u128 avoids overflow when input * fee approaches u64::MAX.
+    // u128 + checked math avoids overflow when input * fee approaches u64::MAX.
     let amm = &ctx.accounts.amm;
-    let fee_amount = ((input as u128) * (amm.fee as u128) / 10000) as u64;
-    let taxed_input = input - fee_amount;
+    let fee_amount = (input as u128).checked_mul(amm.fee as u128).unwrap().checked_div(10000).unwrap() as u64;
+    let taxed_input = input.checked_sub(fee_amount).unwrap();
 
     let pool_a = &ctx.accounts.pool_account_a;
     let pool_b = &ctx.accounts.pool_account_b;
--- a/tokens/token-swap/anchor/programs/token-swap/src/instructions/swap_exact_tokens_for_tokens.rs
+++ b/tokens/token-swap/anchor/programs/token-swap/src/instructions/swap_exact_tokens_for_tokens.rs
@@ -25,9 +25,11 @@ pub fn swap_exact_tokens_for_tokens(
         input_amount
     };
 
-    // Apply trading fee, used to compute the output
+    // Apply trading fee, used to compute the output.
+    // u128 avoids overflow when input * fee approaches u64::MAX.
     let amm = &ctx.accounts.amm;
-    let taxed_input = input - input * amm.fee as u64 / 10000;
+    let fee_amount = ((input as u128) * (amm.fee as u128) / 10000) as u64;
+    let taxed_input = input - fee_amount;
 
     let pool_a = &ctx.accounts.pool_account_a;
     let pool_b = &ctx.accounts.pool_account_b;
@@ -51,8 +53,8 @@ pub fn swap_exact_tokens_for_tokens(
         return err!(TutorialError::OutputTooSmall);
     }
 
-    // Compute the invariant before the trade
-    let invariant = pool_a.amount * pool_b.amount;
+    // Compute the invariant before the trade (u128 to avoid overflow on large pools)
+    let invariant = (pool_a.amount as u128) * (pool_b.amount as u128);
 
     // Transfer tokens to the pool
     let authority_bump = ctx.bumps.pool_authority;
@@ -99,7 +101,7 @@ pub fn swap_exact_tokens_for_tokens(
                 },
                 signer_seeds,
             ),
-            input,
+            output,
         )?;
         token::transfer(
             CpiContext::new(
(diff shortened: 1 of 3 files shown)


tokens/token-swap/anchor/programs/token-swap/src/instructions/swap_exact_tokens_for_tokens.rs
```diff
@@ -25,9 +25,11 @@ pub fn swap_exact_tokens_for_tokens(
         input_amount
     };
 
-    // Apply trading fee, used to compute the output
+    // Apply trading fee, used to compute the output.
+    // u128 avoids overflow when input * fee approaches u64::MAX.
     let amm = &ctx.accounts.amm;
-    let taxed_input = input - input * amm.fee as u64 / 10000;
+    let fee_amount = ((input as u128) * (amm.fee as u128) / 10000) as u64;
```
dev-jodee: should use checked_ math for better handling of overflow

dev-jodee (changes requested):

tokens/token-swap/anchor/programs/token-swap/src/instructions/swap_exact_tokens_for_tokens.rs
```diff
@@ -26,10 +26,10 @@ pub fn swap_exact_tokens_for_tokens(
     };
 
     // Apply trading fee, used to compute the output.
-    // u128 avoids overflow when input * fee approaches u64::MAX.
+    // u128 + checked math avoids overflow when input * fee approaches u64::MAX.
     let amm = &ctx.accounts.amm;
-    let fee_amount = ((input as u128) * (amm.fee as u128) / 10000) as u64;
-    let taxed_input = input - fee_amount;
+    let fee_amount = (input as u128).checked_mul(amm.fee as u128).unwrap().checked_div(10000).unwrap() as u64;
```
dev-jodee: Doing unwrap is dangerous as well, you should map it to a specific error of your program so its clear to the caller what the issue is (map_err)

dev-jodee (changes requested):

tokens/token-swap/anchor/programs/token-swap/src/instructions/swap_exact_tokens_for_tokens.rs
NikkiAung: Good point — fixed in `08e0bea`: added `TutorialError::MathOverflow` and mapped every `checked_mul`/`checked_div`/`checked_sub` in this function via `.ok_or(TutorialError::MathOverflow)?` instead of `.unwrap()`, so an overflow now returns a clear program error instead of panicking. Re-ran the full suite — all 10 pass.

dev-jodee (approved):

PR #693, opened 2026-08-21
security: gate verify_collection behind the collection's original creator

NikkiAung:
## Summary

`tokens/nft-operations` has an NFT-collection-verification forgery vulnerability, present
identically in **both** the `anchor` and `pinocchio` implementations: `verify_collection`
CPIs into Metaplex's `VerifyCollectionV1`/`Verify` instruction, signed by this program's
`[b"authority"]` PDA — the same PDA `create_collection` sets as the collection's Metaplex
`update_authority`. **Neither implementation checks who is allowed to call
`verify_collection`.** Anchor requires `authority: Signer<'info>` but never compares it to
anything; Pinocchio checks only `payer.is_signer()`. The PDA itself is global (fixed seed, no
per-collection component) and signs unconditionally for whoever invokes the instruction.

### The attack

`create_collection` is fully permissionless, and `mint_nft` lets anyone mint an NFT with an
unverified `collection` field pointing at *any* `collection_mint` (freely settable — an
unverified claim requires no permission). So:

1. Attacker mints their own NFT via `mint_nft`, claiming membership in a victim's real,
   already-created collection.
2. Attacker calls `verify_collection`, supplying the victim's public collection accounts and
   their own throwaway signer.
3. The program derives the same global PDA (genuinely the victim collection's real
   `update_authority`) and signs — Metaplex's own checks pass, since from its perspective the
   signer is exactly right. It has no way to know an unrelated caller triggered this.
4. The attacker's fake NFT is now permanently `verified: true` as an authentic member of the
   victim's collection — a real NFT-marketplace forgery primitive, since most marketplaces
   and wallets only trust the verified badge.

**Scoping the *signing* PDA per collection would not fix this** — I checked this carefully
before implementing. PDA derivation is deterministic and public: an attacker supplying the
victim's real `collection_mint` would still cause the program to derive and sign with that
exact collection's genuine authority. The fix has to be an explicit check on the *caller's*
identity, not a change to what signs.

### Why this went unnoticed

- The anchor README (`tokens/nft-operations/anchor/readme.MD`) describes `authority` as:
  *"signer of the transaction. This can be used to restrict the address that can execute the
  verify collection method, by adding constraints"* — a vague hint that constraints *could*
  be added, but none actually are, and the concrete consequence is never spelled out. Much
  softer than this repo's `compression/cnft-vault` README, which explicitly and prominently
  discloses its own analogous gap as an intentional proof-of-concept limitation — I checked
  and ruled that one out as already-disclosed; this one reads differently.
- The pinocchio test file has a comment revealing the author's actual mental model:
  *"Metaplex `Verify` performs strict checks: the signer must be the collection's update
  authority (our PDA)... A successful transaction therefore proves the whole flow is
  correct."* This shows the gap wasn't a deliberate simplification — Metaplex's own checks
  really are correct, but they say nothing about who invoked *this program's* wrapping
  instruction.
- Neither test suite (`anchor/tests/litesvm.test.ts`, `pinocchio/tests/test.ts`) ever called
  `verify_collection` with a different signer than the one who created the collection — zero
  coverage of the vulnerable path.

## Fix

Adds a `collection_authority` PDA (seeded by `collection_mint`) recording the collection's
actual creator at `create_collection` time, checked in `verify_collection`.
`create_collection` stays fully permissionless — the new account only gates verifying
*someone else's* collection. Per-collection scoping is necessary, not just stylistic: a
global tracking account would just relocate the bug (whoever creates the first collection
would become "the" authorized verifier for every collection).

- **Anchor**: new `state.rs` (`CollectionAuthority { creator: Pubkey }`) and `errors.rs`
  (`MintNftError::Unauthorized`), wired in via `init`/`seeds` constraints on both contexts.
- **Pinocchio**: this program has no existing custom-account infrastructure, so the new PDA
  stores just the raw 32-byte creator pubkey (no discriminator — matches the existing
  precedent in `basics/favorites/pinocchio`, safe since the account only ever exists at its
  derived address if this program's own `create_collection` created it there). Since
  Pinocchio gives none of Anchor's automatic owner/length checks, `verify_collection`
  explicitly verifies `collection_authority.owner() == program_id` and the correct data
  length before trusting its contents — dropping that check is exactly the class of gap a
  prior round in this series (#667) found in other Pinocchio ports.

## Test changes

Both test suites gained a negative test: a second, unrelated signer attempts
`verify_collection` on the first signer's collection.

- Anchor: asserts the Anchor custom error code `Unauthorized`.
- Pinocchio: asserts the specific `InstructionErrorCustom { code: 1 }` *and* the paired
  on-chain log message, not just "the transaction failed" — confirmed empirically what
  LiteSVM's error/log shapes actually look like before writing the assertion, rather than
  assuming.

Verified for both implementations: reverted just the authorization check (kept the new
account plumbing so it still compiles), rebuilt, confirmed an outsider's `verify_collection`
call *succeeds* against the unpatched code, reapplied, confirmed it's rejected.

## Out of scope

- `mint_nft`'s unrestricted `collection_mint` parameter — safe by construction, since an
  unverified claim grants no privilege on its own.
- Lack of PDA-seeds constraints on `metadata`/`master_edition`/`collection_metadata` — this
  example consistently leans on Metaplex's own instruction processors to re-derive and
  validate these against the mint; not a gap this fix needs to touch.
- README documentation drift (the embedded code snippets and the vague "can be used to
  restrict..." line will go stale) — worth a follow-up docs-only edit, not done here.

## Verification

- `anchor build --ignore-keys`, full `litesvm.test.ts` suite: 4/4 passing.
- `cargo build-sbf` + pinocchio test suite: 4/4 passing.
- `cargo clippy -- -D warnings` and `cargo fmt --check` clean for the Pinocchio crate (a root
  workspace member, unlike the Anchor crate which is `.workspace-ignore`d).
- `tsc --noEmit` and root `prettier --check` clean for both.

### DIFF
--- a/tokens/nft-operations/anchor/programs/mint-nft/src/contexts/create_collection.rs
+++ b/tokens/nft-operations/anchor/programs/mint-nft/src/contexts/create_collection.rs
@@ -1,3 +1,4 @@
+use crate::state::CollectionAuthority;
 use anchor_lang::prelude::*;
 use anchor_spl::{
     associated_token::AssociatedToken, 
@@ -44,6 +45,16 @@ pub struct CreateCollection<'info> {
     )]
     /// CHECK: This account is not initialized and is being used for signing purposes only
     pub mint_authority: UncheckedAccount<'info>,
+    // Records `user` as this collection's creator, so verify_collection can
+    // later confirm only they (not an arbitrary caller) may verify members.
+    #[account(
+        init,
+        payer = user,
+        space = CollectionAuthority::LEN,
+        seeds = [b"collection_authority", mint.key().as_ref()],
+        bump,
+    )]
+    pub collection_authority: Account<'info, CollectionAuthority>,
     #[account(mut)]
     /// CHECK: This account will be initialized by the metaplex program
     metadata: UncheckedAccount<'info>,
@@ -65,6 +76,7 @@ pub struct CreateCollection<'info> {
 
 impl<'info> CreateCollection<'info> {
     pub fn create_collection(&mut self, bumps: &CreateCollectionBumps) -> Result<()> {
+        self.collection_authority.creator = self.user.key();
 
         let metadata = &self.metadata.to_account_info();
         let master_edition = &self.master_edition.to_account_info();
--- a/tokens/nft-operations/anchor/programs/mint-nft/src/contexts/verify_collection.rs
+++ b/tokens/nft-operations/anchor/programs/mint-nft/src/contexts/verify_collection.rs
@@ -1,3 +1,4 @@
+use crate::{errors::MintNftError, state::CollectionAuthority};
 use anchor_lang::prelude::*;
 
 use anchor_spl::metadata::mpl_token_metadata::instructions::{
@@ -28,6 +29,15 @@ pub struct VerifyCollectionMint<'info> {
     /// CHECK: This account is not initialized and is being used for signing purposes only
     pub mint_authority: UncheckedAccount<'info>,
     pub collection_mint: Account<'info, Mint>,
+    // Only the wallet recorded as this collection's creator at create_collection
+    // time may verify NFTs into it — the mint_authority PDA otherwise signs
+    // unconditionally for whoever calls this instruction.
+    #[account(
+        seeds = [b"collection_authority", collection_mint.key().as_ref()],
+        bump,
+        constraint = collection_authority.creator == authority.key() @ MintNftError::Unauthorized,
+    )]
+    pub collection_authority: Account<'info, CollectionAuthority>,
     #[account(mut)]
     pub collection_metadata: Account<'info, MetadataAccount>,
     pub collection_master_edition: Account<'info, MasterEditionAccount>,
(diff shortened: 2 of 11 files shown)


tokens/nft-operations/anchor/programs/mint-nft/src/contexts/verify_collection.rs
```diff
@@ -28,6 +29,15 @@ pub struct VerifyCollectionMint<'info> {
     /// CHECK: This account is not initialized and is being used for signing purposes only
     pub mint_authority: UncheckedAccount<'info>,
     pub collection_mint: Account<'info, Mint>,
+    // Only the wallet recorded as this collection's creator at create_collection
```
dev-jodee, line 32: can you please remove a lot of the comments are surperfluous / self explaining. For future prs, make sure your agent doesn't add that many comments / trim down comments or remove them if its self explanatory

tokens/nft-operations/anchor/programs/mint-nft/src/contexts/create_collection.rs
dev-jodee, line 43: instead of adding a new account and requiring rent & initialize, you could add the user's pubkey in the mint_authority's seeds

dev-jodee (changes requested):

PR #731, opened 2026-09-11
feat(token-fundraiser): add teardown for failed campaigns

### LINKED ISSUE #725: token-fundraiser: no teardown for a failed campaign (vault + fundraiser rent stranded)
## Summary

The `token-fundraiser` example has no teardown path for a **failed** campaign (one that expires without meeting its target). This affects both implementations:

- `tokens/token-fundraiser/anchor`
- `tokens/token-fundraiser/pinocchio` (added in #708)

## Details

The program exposes four instructions — `initialize`, `contribute`, `check_contributions` (the anchor `checker`), and `refund` — and only tears down accounts on the **success** path:

- **Success:** `check_contributions` requires `vault.amount >= amount_to_raise`, transfers the **full** `vault.amount` to the maker, and closes the vault + fundraiser. Any tokens sent directly into the vault are swept to the maker here, so nothing is stranded on this path.
- **Failure:** once a campaign expires under target, contributors call `refund`, which returns each contributor's **recorded** amount and closes only that contributor's record. There is no instruction that closes the vault or the fundraiser account, so:
  - the vault ATA and fundraiser PDA are left behind with their **rent unrecovered**, and
  - any tokens transferred **directly** into the vault (not via `contribute`, so not part of any recorded contribution) stay **permanently locked**.

This gap is inherited from the reference implementation — the anchor version has the same four instructions and no failed-campaign teardown — so it is not specific to the pinocchio port.

## Why not just gate on the vault balance

Switching `check_contributions` back to releasing based on the raw vault balance is **not** the fix: because anyone can transfer into a standard ATA, that would let a stranger force-release a campaign that never met its target (and block legitimate refunds). That is a more serious issue than stranded rent, so both implementations intentionally gate release on the recorded total.

## Proposed follow-up

Add a maker-callable teardown instruction for a failed campaign — allowed once the duration has elapsed and the target was not met — that sweeps any remaining vault balance back to the maker and closes the vault + fundraiser accounts, recovering their rent. To keep the examples in parity, this should be applied to **both** the anchor and pinocchio implementations together rather than diverging one.

## Context

Raised during review of #708: https://github.com/solana-foundation/program-examples/pull/708#discussion_r3889947500

devtechedge:
## What and why

A failed campaign (one that expires under target) had no teardown path: `refund` returns each recorded contribution and closes the contributor record, but the vault and the fundraiser account stayed behind with their rent unrecovered, and any tokens deposited directly into the vault stayed locked.

This adds a maker-callable `teardown` instruction. It is allowed once the duration has elapsed, requires every recorded contribution to be refunded first (`current_amount == 0`), sweeps the remaining vault balance to the maker, and closes the vault and the fundraiser account.

The refund-first gate matters: without it, sweeping `vault.amount` straight to the maker would also capture tokens that still belong to contributors who have not refunded yet. Once every record is refunded, the only balance left in the vault is stray direct deposits, which the maker can safely collect.

One note on the issue's parity ask: #708 was closed unmerged, so only the anchor implementation exists on main. When a pinocchio port of this example lands, the same instruction should be added there to keep the flavors in sync.

Fixes #725

## Testing

Two cases added to `tests/litesvm.test.ts`:

- teardown is rejected with `UnrefundedContributions` while a recorded contribution is still un-refunded (runs after the deadline warp, before the refund)
- after the refund drains the vault, a stray direct deposit into the vault is swept to the maker and both the vault and the fundraiser accounts are closed

I could not run `anchor test` locally (no SBF toolchain on this machine); the added LiteSVM tests run through the example's `anchor test` in CI.

## AI disclosure

Check exactly one. See [CONTRIBUTING.md](https://github.com/solana-foundation/program-examples/blob/main/CONTRIBUTING.md#ai-use).

- [ ] No AI tooling was used beyond editor autocomplete.
- [x] AI tooling was used. Tool and extent: drafted the new instruction and tests with Claude; I reviewed the full diff against the existing checker/refund instructions and the design notes in the issue before opening.

### DIFF
--- a/tokens/token-fundraiser/anchor/programs/fundraiser/src/error.rs
+++ b/tokens/token-fundraiser/anchor/programs/fundraiser/src/error.rs
@@ -16,6 +16,8 @@ pub enum FundraiserError {
     FundraiserNotEnded,
     #[msg("The fundraiser has ended")]
     FundraiserEnded,
+    #[msg("Contributions have not all been refunded yet")]
+    UnrefundedContributions,
     #[msg("Invalid total amount. i should be bigger than 3")]
     InvalidAmount
 }
--- a/tokens/token-fundraiser/anchor/programs/fundraiser/src/instructions/mod.rs
+++ b/tokens/token-fundraiser/anchor/programs/fundraiser/src/instructions/mod.rs
@@ -2,8 +2,10 @@ pub mod initialize;
 pub mod contribute;
 pub mod checker;
 pub mod refund;
+pub mod teardown;
 
 pub use initialize::*;
 pub use contribute::*;
 pub use checker::*;
 pub use refund::*;
+pub use teardown::*;
--- a/tokens/token-fundraiser/anchor/programs/fundraiser/src/instructions/teardown.rs
+++ b/tokens/token-fundraiser/anchor/programs/fundraiser/src/instructions/teardown.rs
@@ -98,7 +98,7 @@ impl<'info> Teardown<'info> {
             authority: self.fundraiser.to_account_info(),
         };
 
-        close_account(CpiContext::new_with_signer(cpi_program, close_accounts, &signer_seeds), None)?;
+        close_account(CpiContext::new_with_signer(cpi_program, close_accounts, &signer_seeds))?;
 
         Ok(())
     }
--- a/tokens/token-fundraiser/anchor/programs/fundraiser/src/instructions/teardown.rs
+++ b/tokens/token-fundraiser/anchor/programs/fundraiser/src/instructions/teardown.rs
@@ -1,0 +1,105 @@
+use anchor_lang::prelude::*;
+use anchor_spl::{
+    associated_token::AssociatedToken,
+    token::{
+        close_account,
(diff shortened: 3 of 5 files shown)


devtechedge:
Hello - soft check-in on this token-fundraiser teardown whenever you have a moment.

A failed campaign could refund contributors, but the vault and fundraiser accounts stayed behind with their rent unrecovered, and any tokens deposited straight into the vault stayed locked (`Fixes #725`).

This adds a maker-callable `teardown` instruction in `tokens/token-fundraiser/anchor/programs/fundraiser/src/instructions/teardown.rs`, gated on the duration elapsing and `current_amount == 0` so unrefunded contributions cannot be swept to the maker.

Two LiteSVM cases in `tests/litesvm.test.ts` cover the `UnrefundedContributions` rejection and the post-refund stray-deposit sweep that closes both accounts.

CI is green on my side, and I am happy to adjust the gate, rename the error, or mirror the instruction when a pinocchio port lands.

No rush at all - the branch is still ready whenever it fits the queue.

PR #732, opened 2026-09-11
docs(pda-rent-payer): caution note on the unrestricted caller of create_new_account

### LINKED ISSUE #671: basics/pda-rent-payer: create_new_account has no caller check — worth a note for people copying it?
`create_new_account` spends from a program-owned vault, and nothing in the
instruction constrains who may call it.

The accounts struct is:

```rust
pub struct CreateNewAccount<'info> {
    #[account(mut)]
    new_account: Signer<'info>,

    #[account(mut, seeds = [b"rent_vault"], bump)]
    rent_vault: SystemAccount<'info>,
    system_program: Program<'info, System>,
}
```

`new_account` is a `Signer`, but it signs for itself — it is the account being
created, and the caller generates that keypair. `rent_vault` is a PDA on a
single literal seed, shared by everyone. So there is no authority account and
no `has_one`, and each call moves one rent-exempt minimum out of the shared
vault into an account the caller controls.

That is almost certainly intentional for the example's purpose — it exists to
show a PDA signing for itself, and it does that cleanly and readably. My
question is only about the people who copy it, since a rent vault is exactly
the kind of helper that gets lifted into a real program.

Would a short comment on the accounts struct be welcome? Something like:

```rust
// NOTE: this example does not restrict who may call it. A real rent vault
// needs an authority check — a `has_one` against an admin recorded at
// initialisation, seeds that bind the vault to one funder, or a per-caller
// limit. (A bare `authority: Signer` alone is not enough — any keypair
// can sign for itself.)
```

Happy to open a PR with just that comment if it's useful. I did not want to add
a constraint, since that would change what the example teaches.

For context: `basics/checking-accounts` is the example that teaches exactly this
check, so a pointer between the two might do the job on its own.

devtechedge:
## What and why

`create_new_account` spends from a shared `rent_vault` PDA and the only signer is the newly created account itself, so anyone can drain the vault one rent-exempt minimum at a time. That is intentional here - the example exists to show a PDA signing for itself - but it is exactly the kind of helper that gets copied into real programs, per the discussion in #671.

This adds the suggested caution comment above the instruction in all three flavors (anchor, native, pinocchio) so the note travels with whatever version a reader copies. No behavior changes; the example keeps teaching what it teaches.

Fixes #671

## AI disclosure

Check exactly one. See [CONTRIBUTING.md](https://github.com/solana-foundation/program-examples/blob/main/CONTRIBUTING.md#ai-use).

- [ ] No AI tooling was used beyond editor autocomplete.
- [x] AI tooling was used. Tool and extent: drafted the comment text with Claude; I reviewed the diff across the three flavors and matched each file's style before opening.

## Testing

Comment-only change, no code touched.

### DIFF
--- a/basics/pda-rent-payer/anchor/programs/anchor-program-example/src/instructions/create_new_account.rs
+++ b/basics/pda-rent-payer/anchor/programs/anchor-program-example/src/instructions/create_new_account.rs
@@ -1,6 +1,11 @@
 use anchor_lang::prelude::*;
 use anchor_lang::system_program::{create_account, CreateAccount};
 
+// NOTE: this example does not restrict who may call it. A real rent vault
+// needs an authority check: a `has_one` against an admin recorded at
+// initialization, seeds that bind the vault to one funder, or a per-caller
+// limit. A bare `authority: Signer` alone is not enough, since any keypair
+// can sign for itself.
 #[derive(Accounts)]
 pub struct CreateNewAccount<'info> {
     #[account(mut)]
--- a/basics/pda-rent-payer/native/program/src/instructions/create_new_account.rs
+++ b/basics/pda-rent-payer/native/program/src/instructions/create_new_account.rs
@@ -8,6 +8,10 @@ use solana_program::{
 
 use crate::state::RentVault;
 
+// NOTE: this example does not restrict who may call it. A real rent vault
+// needs an authority check: an authority recorded at initialization with
+// seeds that bind the vault to one funder, or a per-caller limit. A bare
+// signer alone is not enough, since any keypair can sign for itself.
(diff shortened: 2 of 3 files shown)


devtechedge:
Hi there - gentle check-in on this pda-rent-payer caution note whenever you have a moment.

`create_new_account` spends from a shared `rent_vault` PDA where the only signer is the newly created account, so anyone can drain the vault one rent-exempt minimum at a time - intentional for the example, but easy to copy into a real program (`Fixes #671`).

This adds the suggested caution comment above the instruction in all three flavors: anchor, native, and pinocchio `create_new_account.rs` files.

There is no behavior change; the example still teaches PDA self-signing, and the note simply travels with whatever flavor a reader copies.

Happy to rephrase the caution, trim it, or scope it to a single flavor if that is preferred.

Whenever it fits the review queue is fine - still ready on my side.

